Issue #82, opened 2016-10-29
on maps, ctrl+clicking an icon opens the link in the same window instead of in a new tab


maxlath:

maxlath:
fixed by #83

PR #83, opened 2016-10-29
Fix for ctrl+click on map (issue #82)

mtbottle:
This is a fix for issue #82 to get the ctrl/cmd map to work. Also added metaKey to work on mac...


app/lib/utils.coffee
```diff
@@ -87,7 +87,8 @@ module.exports = (Backbone, _, $, app, window)->
 
   # anchor with a href are opened out of the current window
   # when the ctrlKey is pressed: the normal action should thus be prevented
-  isOpenedOutside: (e)-> e.ctrlKey
+  isOpenedOutside: (e)->
+    e.ctrlKey or e.metaKey
```
maxlath: one-liner can be on the same line

Issue #73, opened 2016-05-24
elastic textarea keeping its previous height after a message was posted


maxlath:
The incriminated Behavior: [app/modules/general/behaviors/elastic_textarea.coffee](/inventaire/inventaire-client/blob/master/app/modules/general/behaviors/elastic_textarea.coffee)

The desired behavior: reset the textarea when empty.

Problem: this might have to do with [autosize](https://github.com/jackmoore/autosize) itself

![capture d ecran 2016-05-24 a 14 41 32](https://cloud.githubusercontent.com/assets/1596934/15508353/61dfceb2-21cf-11e6-907c-6383fbf46237.png)

maxlath:
closed by 1ea87fd

PR #85, opened 2016-10-29
Fix text box resizing (issue #73)

mtbottle:
This fixes the text box rendering in #73

### DIFF
--- a/app/modules/general/plugins/messages.coffee
+++ b/app/modules/general/plugins/messages.coffee
@@ -28,5 +28,5 @@ module.exports =
     err.selector = '.alertBox'
     forms_.alert(@, err)
 
-  emptyTextarea: -> @ui.message.val('')
+  emptyTextarea: -> @render()
   recoverMessage: (message)-> @ui.message.val message


maxlath:
so it appears that in the case the view has registered to the [BackupForm behavior](https://github.com/inventaire/inventaire-client/blob/0bd871ab969104ec6af2215296e09f130e8730d3/app/modules/general/behaviors/backup_form.coffee) (see [Marionette documentation](http://marionettejs.com/docs/v2.4.3/marionette.behaviors.html) if you want to learn more on what they call a Behavior), as this is the case for the [transaction view](https://github.com/inventaire/inventaire-client/blob/0bd871ab969104ec6af2215296e09f130e8730d3/app/modules/transactions/views/transaction.coffee), re-rendering the view isn't enough to empty the textarea as the goal of the BackupForm behavior is precisely to preserve text from being lost on a view re-render: emptying the textarea before re-rendering seems like the way to go.

To test the transaction view: create two users, one with a book to give share or sell with visibility mode public, and make a request from the other user account. This should open a transaction discussion visible at [http://localhost:3006/transactions](http://localhost:3006/transactions), using the [transaction view](https://github.com/inventaire/inventaire-client/blob/0bd871ab969104ec6af2215296e09f130e8730d3/app/modules/transactions/views/transaction.coffee) mentioned above

maxlath:
finally merged as 1ea87fd, sorry for the delay :)

Issue #89, opened 2016-11-30
integrating RSS feeds


maxlath:
Thank to @jum-s, [the server will now expose RSS feeds](https://github.com/inventaire/inventaire/issues/71), this issue discuss how to make this feature known from the client.

### minimal integration
Add a link in the `<head>`:
```html
<link rel="alternate" type="application/atom+xml" href="/feeds/url">
```
This should let browsers discover those feeds.
See [app/modules/general/lib/head_metadata.coffee](https://github.com/inventaire/inventaire-client/blob/943fe11140ade7a0f966aebbc9097d9d974e3cf8/app/modules/general/lib/head_metadata.coffee) for implementation

### visible elements
This should probably be something discreet as its rather an expert user feature. A simple text link at the end of the page à la Github?
![capture du 2016-11-30 17-14-09](https://cloud.githubusercontent.com/assets/1596934/20760502/79359088-b720-11e6-9ad1-fed5d1f81436.png)
Possible problem with this option: infinite scroll pages don't really have a footer

maxlath:
how goodreads does it
![capture du 2017-01-12 19-24-34](https://cloud.githubusercontent.com/assets/1596934/21902607/d7221448-d8fc-11e6-879d-b24e42e53c24.png)

maxlath:
@jum-s integration in progress!

![profile](https://cloud.githubusercontent.com/assets/1596934/23574098/67b4ef1c-007b-11e7-8e7f-cb83f5ed72ad.png)

PR #94, opened 2017-02-09
add atom link as metadata in index.html

jum-s:
looks simple when its working !


maxlath:
associated issue: #89

maxlath:
I have been playing a bit with history, which seem to have messed up this PR: there should only be  	838c926 and 9f1a17c referenced here.
plouf-mini

maxlath:
merged!

PR #140, opened 2018-03-03
Isbns importer

maxlath:
still todo:
- find an elegant way to deal with editions that can't be found anywhere


maxlath:
finally done! @jum-s sorry for the big PR, when can test/review it together at your convenience

Issue #144, opened 2018-04-15
clic on work image links to the zoomed image


jum-s:
When I click on an work image from an author page, it opens a zoom on the image, 
 while i would like to go to the entity description page, not only having a bigger image.

maxlath:
so indeed that's a bug, but I still would like to have this possibility to zoom on the image by clicking the magnifying glass icon

maxlath:
fixed by #145

PR #145, opened 2018-04-15
replace work image zoom by a link to the work

jum-s:
from issue #144


maxlath:
so indeed that's a bug, but I still would like to have this possibility to zoom on the image by clicking the     magnifying glass icon

PR #176, opened 2019-05-05
Small english grammar fix that makes a big difference :)

trosel:


maxlath:
thanks! :D
I often wonder how obvious it is that most of the English texts were written by non-native speakers >< If you spot more mistakes, you are very welcome to open more PRs!

PR #190, opened 2020-01-03
update node to v12

maxlath:
to be aligned with the server (see https://github.com/inventaire/inventaire/pull/351)
 
Required to use a non-published version of brunch-sass
as the latest published version was depending on an old node-sass version
which was crashing


jum-s:
sure thing, lets do it

Issue #191, opened 2020-01-10
[item view] align remove button


jum-s:
Delete item button is not aligned with other elements in the view, making it harder to see

![Sélection_010](https://user-images.githubusercontent.com/5363918/72176309-064dc780-33d6-11ea-8d72-c564208f8ea2.jpg)

 A better view woould align it with the rest of the elements, such as : 

![Sélection_009](https://user-images.githubusercontent.com/5363918/72176366-1fef0f00-33d6-11ea-8ee9-07d9008f2a6e.jpg)

maxlath:
fixed by #192

PR #192, opened 2020-01-10
align delete button in item view with other elements

jum-s:
also uniformize vertical seperation and margin around
block ".one" and block ".two"

Problem : now `#itemData` element is having a bottom margin way to large (as seen from issue #191  screenshot), this is due to an inline rule inserted on the dom element : `<div id="itemData" class="two" style="min-height: 367px;">`. This rule override any scss rule and should be delete or changed to `350px` to reduce the bottom margin, but i dont really know where this inline rule come from...


maxlath:
@jum-s the inline rule was added by `item_show_data`, I removed it (85351a5) as that was indeed quite hacky and possibly not needed anymore

PR #196, opened 2020-02-26
extract itemsData out of InventoryBrowser

jum-s:
to allow other consumers than user inventory to display an InventoryBrowser

### DIFF
--- a/app/modules/inventory/views/inventory_browser.coffee
+++ b/app/modules/inventory/views/inventory_browser.coffee
@@ -52,9 +52,11 @@ module.exports = Marionette.LayoutView.extend
 
   initBrowser: ->
     startLoading.call @, { selector: '#browserFilters', timeout: 180 }
-    { itemsData } = @options
+    { itemsDataPromise } = @options
     @spreadData itemsData
-    waitForInventoryData = @showItemsListByIds(@itemsByDate)
+    waitForInventoryData = itemsDataPromise
+      .then @ifViewIsIntact('spreadData')
+      .then @ifViewIsIntact('showItemsListByIds', null)
 
     waitForEntitiesSelectors = waitForInventoryData
       .then @ifViewIsIntact('showEntitySelectors')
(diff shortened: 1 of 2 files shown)


maxlath:
@jum-s  I see one issue with the proposed implementation: if the `inventory-view` endpoints takes 20 seconds to respond (which is the case today for large inventories in absence of [caching](https://github.com/inventaire/inventaire/pull/368), and even after that, that will be the case on cache misses), nothing is shown, not even the previously displayed loading spinner and controls (allowing to switch display mode for instance)

(This situation can be demonstrated by adding `.delay 10000` between [those 2 lines](https://github.com/inventaire/inventaire-client/blob/335a4c3/app/modules/inventory/views/inventory_layout.coffee#L108-L109))

A workaround could be to call `InventoryBrowser` earlier, with only a promise of `itemsData`:
in `InventoryLayout` `showInventoryBrowser` method
```coffee
_.preq.get app.API.items.inventoryView(params)
.then (itemsData)=>
  @itemsList.show new InventoryBrowser { itemsData, model }
```
would become
```coffee
itemsDataPromise = _.preq.get app.API.items.inventoryView(params)
@itemsList.show new InventoryBrowser { model, itemsDataPromise }
```
then, in `InventoryBrowser` `initBrowser`
```coffee
{ itemsDataPromise } = @options
waitForInventoryData = itemsDataPromise
  .then @ifViewIsIntact('spreadData')
  .then @ifViewIsIntact('showItemsListByIds', null)
```

app/modules/inventory/views/inventory_browser.coffee
```diff
@@ -51,9 +51,11 @@ module.exports = Marionette.LayoutView.extend
   onShow: -> @initBrowser()
 
   initBrowser: ->
+    { itemsDataPromise } = @options
```
maxlath: `itemsDataPromise` isn't defined on `@options`, `getItemsDataPromise` is

app/modules/inventory/views/inventory_browser.coffee
```diff
@@ -51,9 +51,11 @@ module.exports = Marionette.LayoutView.extend
...
+    @spreadData itemsData
```
maxlath:
```suggestion
```

PR #199, opened 2020-03-06
enable cursor select entity data in edition modal

jum-s:
As the item modal is the quickest way to access an isbn in my inventory, I would like to select the ISBN text (and title etc) to copy paste it wherever i want. 
Sélection_019
This solution enable text selection at the price of a smaller clickable area (only the text and the image)

### DIFF
--- a/app/modules/inventory/scss/_item_show.scss
+++ b/app/modules/inventory/scss/_item_show.scss
@@ -8,11 +8,11 @@
     max-width: 36em;
     min-height: 350px;
   }
-  a.edition, a.work, .author-preview a, .series-preview a{
+  .edition, .work, .author-preview a, .series-preview a{
     margin: 0.2em 0;
     @include bg-hover(#f3f3f3);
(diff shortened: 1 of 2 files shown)


maxlath:
I'm uncomfortable with this change as it breaks the affordance of "this is some data about the edition, you can click me to know more as indicated by the background color change on hover".  I guess it could be acceptable if the ISBN was out of the hover zone though.

Alternatively, there are tricks to get the ISBN in other ways:
* right click the link and select "Copy the link" will give you the edition URL which contains the ISBN in it's non-hyphenated form (this trick will break the day we [stop using ISBNs as canonical URIs](https://github.com/inventaire/inventaire/issues/192) though)
* start selecting below the link and drag the mouse up in order to select the ISBN, then press Ctrl+C

While we are here, would that be too much if we were to display the edition publisher (P123) and the work subject (P921) on that item view? maybe only on large screens?

PR #205, opened 2020-05-03
Entities table lists

jum-s:
Before going further with other entities lists, a validation that works lists table is on an "ok" path would be nice


maxlath:
A re-styling of entities list is long overdue, so yes, that's a definitely an ok path! \o/

Having a "display" settings section could make sense if the display settings are hard to reach from the interface they control, which could be the case on small screen where we might sacrifice "display mode" buttons.

Regarding the current implementation
#### custom template
you can use the view [`getTemplate` method](https://marionettejs.com/docs/v2.4.1/marionette.view.html#change-which-template-is-rendered-for-a-view):
```coffee
  getTemplate: ->
    display = localStorageProxy.getItem('entities:display') or 'cascade'
    if display is 'table' then require './templates/work_li_table'
    else require './templates/work_li'
```

#### settings
Storing data in DOM attributes for a unique event dispatcher to consume is fine, I would nevertheless recommend to use [data attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes): `<a data-display-type="table">` should then be accessible with `e.currentTarget.dataset['display-table']`

jum-s:
Is this data attribute thinggy something blocking to merge this PR ? 
i tried to implement it but cant grasp how this attribute will be kept over different sessions (if pressing F5). 
Then i dropped it since something else (`localStorageProxy`) is already doing the job well for other settings.

app/modules/entities/views/work_li.js
```diff
@@ -1,23 +1,36 @@
...
+import { localStorageProxy } from 'lib/local_storage'
 
 export default Marionette.ItemView.extend({
-  template: workLiTemplate,
   className () {
     const prefix = this.model.get('prefix')
     if (this.wrap == null) this.wrap = this.options.wrap
     const wrap = this.wrap ? 'wrapped wrappable' : ''
-    return `workLi entity-prefix-${prefix} ${wrap}`
+    const workLiDisplay = this.display === 'entitiesTable' ? 'workLiTable' : 'workLi'
```
maxlath, line 13: bad news: the `initialize` function is called after `className` (thus the convention to always define `className` above, to try to give the hint/reminder), so `this.display` will always be undefined at this point

app/modules/entities/views/work_li.js
```diff
@@ -1,23 +1,36 @@
...
+      return workLiTemplateTable
+    } else {
+      return workLiTemplate
+    }
+  },
+
   initialize () {
     let preventRerender
+
+    this.display = localStorageProxy.getItem('entitiesDisplay') || 'entitiesCascade'
```
maxlath, line 33: this `entitiesDisplay` local storage key is inconsistent with the `entities:display`, so one or the other is wrong

app/modules/settings/views/templates/display_settings.hbs
```diff
@@ -0,0 +1,35 @@
...
+
+<section class="entities">
+  <h3>{{i18n 'works'}} / {{i18n 'series'}} / {{i18n 'authors'}} {{i18n 'lists'}}</h3>
+  <a class="selector">
+    <span class="label">{{I18n 'current display'}} : </span>
+    <div class="{{#if entitiesCascade}}shown{{/if}}">{{icon 'th-large'}} {{I18n 'cascade'}}</div>
+    <div class="{{#if entitiesTable}}shown{{/if}}">{{icon 'align-justify'}} {{I18n 'table'}}</div>
+  </a>
+  <ul id="entitiesDisplayOptions">
+    <li displayType="entitiesDisplay" id="entitiesCascade">
```
maxlath, line 11: as that's a non standard attribute, it should be define with the [`data` prefix](https://developer.mozilla.org/fr/docs/Web/HTML/Global_attributes/data-*)
```suggestion
    <li data-display-type="entitiesDisplay" id="entitiesCascade">
```

same for the 3 other occurrences

app/modules/settings/views/templates/display_settings.hbs
```diff
@@ -0,0 +1,35 @@
...
+    <div class="{{#if inventoryTable}}shown{{/if}}">{{icon 'align-justify'}} {{I18n 'table'}}</div>
+  </a>
+  <ul id="inventoryDisplayOptions">
+    <li displayType="inventoryDisplay" id="inventoryCascade">
+      <a>{{icon 'th-large'}} {{I18n 'select cascade display'}}</a>
+    </li>
+    <li displayType="inventoryDisplay" id="inventoryTable">
+      <a>{{icon 'align-justify'}} {{I18n 'select table display'}}</a>
+    </li>
+  </ul>
```
maxlath, line 34: part of the reason my review on this PR was lagging is that this menu does not look very clear to me, and could be simplified (from the user point of view) by behaving like a `<select>` and having the selector and the information on the current state of that selector together (like the display selector in the inventory browser)

or even much simpler, an actual `<select>` (but then we loose the icons, but maybe that's ok for now as all this will be replaced by Svelte soon, which will make it trivial ;))
```suggestion
  <label for="inventoryDisplaySelector">{{i18n 'inventory lists'}}</label>
  <select id="inventoryDisplaySelector">
    <option value="inventoryCascade" {{#if inventoryCascade}}selected{{/if}}>{{I18n 'cascade'}}</option>
    <option value="inventoryTable" {{#if inventoryTable}}selected{{/if}}>{{I18n 'table'}}</option>
  </select>
```

maxlath (changes requested):

app/modules/settings/views/templates/display_settings.hbs
```diff
@@ -0,0 +1,35 @@
+<h2 class="layout-title">{{I18n 'display'}}</h2>
```
maxlath, line 1: this appears as `afficher` in French, while that should probably be `affichage`, but as there is no distinction in English, we need to use a trick to make the distinction, which is to set different keys. In the past, we have been using `x_verb` (ex: `search_verb`), so either we could replace previous occurrences of `display` by `display_verb` or replace that one by `display_name`

maxlath (approved):
I added a few commits to fill the last holes. I would thus approve after that, but the reverse review is now needed.

jum-s:
great !
before merge, i think this this work should either be extended to other kind of lists (namely series and authors), or [settings labels](https://github.com/inventaire/inventaire-client/blob/37faf05101e1fba0e5fde6ecb7697d5168d32725/app/modules/settings/views/templates/display_settings.hbs#L4) shall be modified to only `works`

PR #209, opened 2020-05-24
Move entity to wd

jum-s:
Allow to push a specific P31 per type
Based on the `data?action=aliases` endpoint written in https://github.com/inventaire/inventaire/pull/401

maybe a `default` label next to `Q571` and `Q5` would help user choice, but assuming people who push on wikidata are aware of the wd data structure.


jum-s:
close and replace it with #218

PR #212, opened 2020-07-11
Replace exists or create by resolver

maxlath:
depends on https://github.com/inventaire/inventaire/pull/379

started as a simple replacement of endpoint, ended up being much more (see commit messages)

### DIFF
--- a/app/modules/inventory/collections/candidates.coffee
+++ b/app/modules/inventory/collections/candidates.coffee
@@ -1,3 +1,5 @@
+{ looksLikeAnIsbn, normalizeIsbn } = require 'lib/isbn'
+
 module.exports = Backbone.Collection.extend
   model: require '../models/candidate'
 
@@ -10,8 +12,9 @@ module.exports = Backbone.Collection.extend
   addNewCandidates: (newCandidates)->
     alreadyAddedIsbns = @pluck 'normalizedIsbn'
     remainingCandidates = newCandidates.filter isNewCandidate(alreadyAddedIsbns)
-    if remainingCandidates.length > 0 then return @add remainingCandidates
-    else return []
+    if remainingCandidates.length is 0 then return Promise.resolve []
+    addExistingInstancesCounts remainingCandidates
+    .then @add.bind(@)
 
 isSelected = (model)-> model.get('selected')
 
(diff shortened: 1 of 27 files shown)


app/modules/inventory/inventory.coffee
```diff
@@ -101,15 +102,22 @@ showItemsFromModels = (items)->
...
+    when 1
+      item = items.models[0]
+      fallback = -> API.showUserInventory item.get('owner'), true
+      showItemModal item, fallback
     else showItemsList items
 
-showInventory = (options = {})->
+showInventory = (options)->
+  { user, group, section } = options
+  unless user? or group? or section?
```
jum-s, line 113: i guess this will break with shelves merging (and will have to be updated anytime a new category comes in). I understand early return are nice (but maybe there are other reasons) but why not trusting the inventoryLayout to figure out itself its options at initialisation instead ?

app/modules/inventory/lib/parsers/common.coffee
```diff
@@ -7,28 +9,29 @@ module.exports = (data)->
...
   .filter isFirstOccurence(authors)
-  .join ', '
 
 containsNoComma = (str)-> str? and not /,/.test(str)
 
 isFirstOccurence = (array)-> (str, index)->
   noramlizedStr = normalize str
   for previousStr in array[0...index]
-    if noramlizedStr is normalize(previousStr) then return false
+    if leven(noramlizedStr, normalize(previousStr)) < 3 then return false
```
jum-s, line 28: does this number have any reason behind its definition (aka why 3?), maybe would be nice to comment this line with the explanation ?

app/modules/inventory/collections/candidates.coffee
```diff
@@ -21,3 +24,25 @@ setSelected = (bool)-> (model)->
 isNewCandidate = (alreadyAddedIsbns)-> (candidateData)->
   unless candidateData.title? or candidateData.normalizedIsbn? then return false
   return candidateData.normalizedIsbn not in alreadyAddedIsbns
+
+addExistingInstancesCounts = (candidates)->
```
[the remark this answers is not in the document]
maxlath: proposed renaming from 'existing instances' to 'existing entity items' done in 5594bb0f
does it work better?

app/modules/inventory/inventory.coffee
maxlath: that was probably convenient for development then, but doesn't cost much to remove: done in dd5aac42

app/modules/inventory/lib/parsers/common.coffee
maxlath: arbitrary number documented in dacabf82

jum-s (approved):
To me, this indeed achieve parity of features with `exists-or-create-from-seed`
TODO : use enrich to feed publishers labels to the resolver

PR #218, opened 2020-08-24
Add 'instance of' to edit form

jum-s:
autocomplete on matching default suggestions
set default suggestions based on suggestions/wdt_P31.coffee which fetch aliases on
"/api/data?action=aliases"


jum-s:
TODO: find a way to display "recommended" and/or "depreciated" labels inside some autocomplete suggestions

jum-s:
to adress the TODO above, it looks like having a `recommended` flag would necessarily mean some future maintainability to keep the list updated. While `depreciated` feels more stable.
The server could filter depreciated values by default when serving the enpoint

jum-s:
this PR depends on the server branch `aliases-endpoint`

jum-s:
> TODO: find a way to display "recommended" and/or "depreciated" labels inside some autocomplete suggestions

depreciated ids are filter on the server side [17a487e](https://github.com/inventaire/inventaire/pull/416/commits/17a487e50af6e9b73e19bdc9904408f06a64fe24)

maxlath (approved):

maxlath:
@jum-s ping for review after I pushed some commits (I apparently can't request your review back after you asked for mine)

app/modules/entities/views/entities_list_adder.js
```diff
@@ -74,6 +74,12 @@ export default Marionette.CompositeView.extend({
       claims['wdt:P123'] = parentModel.get('claims.wdt:P123')
     }
 
+    if (parentModel.get('claims.wdt:P31')?.length === 1) {
+      const parentP31 = parentModel.get('claims.wdt:P31.0')
```
jum-s, line 78: I would be for reducing to a maximum extent the use of those oneliners `?`, as, to me, they hide complexity by easing access to deep objects

```suggestion
    const P31values = parentModel.get('claims.wdt:P31')
    if (P31values && P31values.length === 1) {
      const parentP31 = P31values[0]
```

app/modules/entities/views/entities_list_adder.js
```diff
@@ -74,6 +74,12 @@ export default Marionette.CompositeView.extend({
       claims['wdt:P123'] = parentModel.get('claims.wdt:P123')
     }
 
+    if (parentModel.get('claims.wdt:P31')?.length === 1) {
```
jum-s, line 77: could we extract that in a named function to be more explicit about whats going on here ?

app/modules/entities/views/entities_list_adder.js
maxlath: I guess it depends of the cases: here, your suggestion does seem clearer (correction made in 8beb8c0c), but we have been using `?` in coffeescript for years and were quite happy about it (it was even listed as one of the [reasons to stay with coffee](https://github.com/inventaire/inventaire/issues/72)); so I wouldn't agree to make it a general rule

app/modules/entities/views/entities_list_adder.js
maxlath: attempted in 8beb8c0c

app/modules/entities/views/entities_list_adder.js
jum-s: Its not only complexity, its also prone to inconsistencies and silent errors. Can we say at least agree that we cannot use this operator if the result is stored in a variable (so a `if` like this one is fine) ?

jum-s:
The more i think about it the less im convinced this is a good feature, do users care so much about what type an entity is that it should be the first thing to see when editing any entity ? Maybe we could keep it to the bottom of the form (revert `b394aa1` )?

maxlath:
I tried a different option in eea20fde:
- prominently showing P31 only to users with dataadmin access level
- still allowing other users to see it after clicking "add more details", and letting P31 above P136, as those 2 properties need to be thought about together (in some way not yet totally clear, https://github.com/inventaire/inventaire/issues/316)

app/modules/entities/views/entities_list_adder.js
maxlath: a good case provided by the server decaffeination:
```coffee
publicationYear = entity.claims['wdt:P577']?[0]?.split('-')[0]
```
was decaffeinated as
```js
const publicationDateClaim = claims['wdt:P577'] && claims['wdt:P577'][0]
const publicationYear = publicationDateClaim && publicationDateClaim.split('-')[0]
```
but once we update the server engine to NodeJS >= 14, we could rewrite it like this
```js
const publicationYear = entity.claims['wdt:P577']?.[0]?.split('-')[0]
```
I do find it more readable

app/modules/entities/views/editor/properties_editor.js
```diff
@@ -36,3 +36,13 @@ export default Marionette.CompositeView.extend({
     this.ui.showAllProperties.hide()
   }
 })
+
+const dataadminOnlyShortlistedProperties = [
+  'wdt:P31'
+]
+
+const getPropertiesShortlistPerRole = propertiesShortlist => {
```
jum-s, line 44: i think by role, you mean access level, since admin too  will have access to this property

app/modules/entities/views/entities_list_adder.js
jum-s: Sure its more readable, no doubt on that. Im saying that this operator *can* make results inconsistent and throw silent errors, which is much more problematic that its readability imo
https://dev.to/macsikora/what-is-wrong-with-optional-chaining-and-how-to-fix-it-3nno

jum-s:
depreciated in favor of  #402

PR #219, opened 2020-09-03
disable piwik/matomo cookies for all users

maxlath:
Analytics cookies aren't strictly required to provide the service and thus should require, under GDPR, to request the user's consent. It's easier to just get rid of them rather than to implement some of those annoying cookie consent box. The [impact on analytics accuracy](https://matomo.org/faq/general/faq_156/) seems very acceptable, given our use case.

Next action: update [privacy statement](https://wiki.inventaire.io/wiki/Privacy)


jum-s (approved):

jum-s:
[wiki](https://wiki.inventaire.io/wiki/Privacy?lang=en) updated.
next action: update the new mediawiki version

Issue #201, opened 2020-03-25
choose a more recent framework to replace Backbone.Marionette+Handlebars


maxlath:
I don't think this is exactly urgent, but 

Good reasons to stop waiting:
* we might be missing out on some serious workflow benefits [to be listed]
* the more we wait, the more code will need to be converted

Good reasons to wait more:
* the more we wait, the less we will be affected by breaking changes going on the meantime (ex: VueJS v3): how many generations of framework can be spare to learn, by just keep doing what we know :D
* #200 needs to be done first (doing both transitions at once doesn't seem like a good idea)

Inspiration to go through this process:
* [RFC: Adopt a modern JavaScript framework for use with MediaWiki](https://phabricator.wikimedia.org/T241180) (summary: [Wikimedia developers dismiss React for JavaScript makeover](https://www.theregister.co.uk/2020/03/20/wikimedia_react_javascript/))

maxlath:
VueJS growingly seems like the right option:
* getting mature+strong community
* Wikimedia community committed to use it
* part of our team is already familiar with it

proposed steps:
* replace brunch with parcel + convert coffeescript to ES6+ #200 
* once [VueJS v3](https://madewithvuejs.com/blog/vue-3-roundup) is out, start playing with it, possibly producing some layouts with it within the client: the progressive integration is one of its advertised features, and that would give the opportunity to change our mind in the process, without having to pay the cost of a full rewrite. This experimentation should, in theory, be quite cheap, thank to VueJS+parcel out-of-the-box integration

maxlath:
or maybe [Svelte](https://svelte.dev/), here is some good propaganda: [why svelte](https://github.com/feltcoop/why-svelte)

maxlath:
[React, Vue, and Svelte Templates Side By Side](https://medium.com/better-programming/react-vue-and-svelte-templates-side-by-side-4aa52cf3cf2)

maxlath:
So Svelte it is! New layouts are now systematically written in Svelte, and some Backbone.Marionette+Handlebars layouts are being rewritten in Svelte too. Once all Backbone.Marionette+Handlebars layouts have been rewritten, the question of the app router and event bus will remain, but that's a matter for another issue

Issue #200, opened 2020-03-25
from coffeescript to ESx


maxlath:
the client-side equivalent to https://github.com/inventaire/inventaire/issues/72

client-side specificities:
* depending on which browsers we want to support, we might need to keep the possibility to transpile to some older version of ES (ES5 ?), which might mean using [Babel](https://babeljs.io/)
* optimize for tree shaking: use ES6 modules?

depending how much work on the bundler setup this conversion involves, that might also be the right time to switch bundler: something like [Parcel](https://parceljs.org/) would for instance bring several benefits over Brunch (mostly out-of-the-box tree shaking, and module splitting)

maxlath:
decaffeination in progress on https://github.com/inventaire/inventaire-client/tree/decaffeinate

maxlath:
#229 merged, we are now fully decaffeinated \o/

PR #229, opened 2020-10-25
CoffeeScript=>ES6 // brunch=>webpack // CommonJS=>ESM

maxlath:
that's quite a big PR ^^'

## Converting CoffeeScript to ES6
addressing #200 

## Changing bundler: moving from brunch to webpack
Brunch has been a great tool, but is now a bit dated. It especially lacks the following features:
- asynchronous module loading / code splitting (as stated in https://brunch.io/docs/why-brunch)
- [tree shaking](https://webpack.js.org/guides/tree-shaking/)

There was an attempt to use [Parcel](https://parceljs.org/) instead, but it had hard to ignore shortcomings:
- no source maps for handlebars and svelte components
- some weird bugs [1](https://github.com/inventaire/inventaire-client/blob/60eb073ed/troubleshooting.md#uncaught-syntaxerror-missing--after-property-list), [2](https://github.com/inventaire/inventaire-client/blob/60eb073ed/troubleshooting.md#--cannot-read-property-type-of-undefined)

So another attempt to setup a new bundler was done with [Webpack](https://webpack.js.org/) instead, which, after reading the [introduction guide](https://webpack.js.org/guides/) is a less of a frightening beast than anticipated, and comes with some serious pros:
- abundant documentation
- large module ecosystem (even if periodically disrupted by new major versions)

With proper cache setup, and after having splitting the app code in smaller chunks (using [dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports)), the time to rebundle in development seems acceptable, and could eventually be further reduce by the use of [snowpack](https://www.snowpack.dev/) (experiments to use it didn't prove fruitful so far).

Good documentation on Webpack:
- https://webpack.js.org/concepts/
- https://webpack.js.org/guides/
- https://webpack.js.org/configuration/
- https://survivejs.com/webpack/preface/

## Changing import syntax: from CommonJS to ESM
ESM is the future: that's the [module mechanism](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) that modern browsers speak, giving the hope that one day we might just get rid of bundlers, if not in production, at least in development (see [snowpack](https://www.snowpack.dev/)). Tools can also rely on its syntax to:
* get better [tree shaking](https://webpack.js.org/guides/tree-shaking/)
* display better code context(?)

As it's the future, using it in the present can be a bit painful: [NodeJS only supports it without an `experimental` flag since v14](https://nodejs.org/api/esm.html#esm_modules_ecmascript_modules)

## Introducing Svelte
In an effort to address #201, this PR includes a [Svelte](https://svelte.dev) setup aside from the soon-to-be-considered-legacy jQuery+Backbone.Marionette+Handlebars. This setup can be tested by going to `/svelte-test` on your development server


jum-s (commented):
~~the exceptional length of this pr implies I cannot really approve this review, it works fine on my setup~~

jum-s:
bug report:

- [ ] in `_item_commons.scss` `.itemBox` dont seem to be called, which gives a broken cascade inventory view
- [ ] when editing a transaction on a item view, item is successfully updated but the view does not refresh, leaving the wrong former icon until another change on the item is done (ie. details update)
- [ ] at `editable_entity.js#L48 avec un `this.invalidateRelationsCache` is not a function
- [ ] when displaying a serie, at `/app/modules/entities/lib/types/serie.js#L77` `this` is undefined (something with `initPartsCollections`)

maxlath:
1. addressed in ff63ac65
2. addressed in 14b0e46e (doesn't work in prod either)
3. could not reproduce
4. addressed in 55c15e1f

Issue #213, opened 2020-08-12
replace entity editor date picker by standard date input


maxlath:
see https://developer.mozilla.org/fr/docs/Web/HTML/Element/input/date

regarding [browsers that aren't implementing the standard yet (Safari)](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#Browser_compatibility), we could either wait for them, or let them have the current implementation, but that would require some mean of feature detection

PR #237, opened 2020-11-13
Live search section

jum-s:
makes a clear separation between searching for entities and searching for groups or users
to be merged after merging server branch `elasticsearch-upgrade`


maxlath:
The behavior to add with click/remove with click combine with a new search on each add/remove, triggering a new loader and thus the whole selection bar to jump is a bit frustrating:
search_categories 2

The jump could be avoided by having the nav at the top, but even then having to click twice when you want to search only one category (which I assume is the most common case) remains frustrating. A solution could instead be to allow only one category at a time. There could be a general "entity" section, which would allow to search all at once

Additionally, a few things broke:
- keyboard navigation doesn't have any effects
key_nav
- subject search crashes as the search is done on `/api/search` instead of Wikidata API as was previously the case

jum-s:
Assuming most people understand the current terms used (authors, works etc) is ok, but i dont think its the case of 'entities'.
Also search in several categories is a feature than has its interest, ie. looking for books of an author, by typing the name of an author in the book category result. Authors books is the most obvious one but of mix can have interesting results in my opinion at least until [#213](https://github.com/inventaire/inventaire/issues/213) is solved
Other possible solutions: 
- a css fixed height corresponding to the amount of maximum result already displayed
- the default could be no selection at all, which then would need only one click to search by category

jum-s:
oh about the keyboard shortcuts, the "nextcategory" and "previous category" have to kindly leave the room if we continue with this feature.

maxlath:
I would be in favor of the one click option, and keep more advance settings for the future advanced search layout #9

jum-s:
so i dont think there is any point of going forward with that pr

maxlath:
closing this PR would mean adapting the `elasticsearch-upgrade` branch to not break the current client implementation where one can query `https://inventaire.io/api/search?types=works|humans|series|publishers|collections|users|groups&search=bla&lang=fr`, right?

jum-s:
yes, meaning `elasticsearch-upgrade` cannot accept an "all" section anymore. 2 possible solutions
- remove all without replacement, keep the bar unified (unlike this pr proposal) 
- split between group and entities sections but add a "all" section in the entities section (disadvantage: its hardly understandable that "all" only concerns the first raw/entities sections)

jum-s:
in any case this pr can be closed, and reopen a new one to solve the aforementioned problem

PR #239, opened 2020-11-18
replace /entity/changes by a generalized version of the contributions layout

maxlath:
depends on https://github.com/inventaire/inventaire/pull/459

## before
Sélection_052

## after
Sélection_051


app/modules/users/scss/contribution.scss
```diff
@@ -62,10 +72,25 @@
...
   .time{
     color: $grey;
-    margin-left: 0.5em;
-    width: 7em;
+    margin-left: auto;
+    padding-left: 0.2em;
+    width: 6em;
     text-align: right;
   }
   .togglers span{
```
jum-s, line 100: nice!
how about modifying `.operation-type` right below to have some more accessible label

from



to something like

app/modules/users/views/templates/contribution.hbs
```diff
@@ -25,6 +29,12 @@
       </div>
     {{/with}}
   </div>
+  {{#if showUser}}
+    <a class="user showUserContributions" href="{{user.contributions}}">
+      <img src="{{imgSrc user.picture 32}}" alt="{{user.username}}">
+      <p class="username">{{user.username}}</p>
+    </a>
+  {{/if}}
   <p class="time" title="{{localTimeString timestamp}}">{{timeFromNow timestamp}}</p>
```
jum-s, line 38: Since this is mainly for log changes, i would be in favor of a more precise date format

manual suggestion (did the gh feature disappear?):

```hbs
<p class="time" title="{{timeFromNow timestamp}}">{{localTimeString timestamp}}</p>
```

maxlath: suggestion still here:
```suggestion
  <p class="time" title="{{timeFromNow timestamp}}">{{localTimeString timestamp}}</p>
```

app/modules/users/scss/contribution.scss
maxlath: better? 1c585b5c

Sélection_053

maxlath: or with a white background
Sélection_054

app/modules/users/views/templates/contribution.hbs
maxlath: done in 0d39c0a5

app/modules/users/scss/contribution.scss
jum-s: nice, either way

PR #246, opened 2020-12-11
live search sections

maxlath:
New attempt to makes a clear separation between searching for entities and searching for groups or users.
To be merged after merging server branch `elasticsearch-upgrade`

Here is a glimpse
multi_search_bar

Pros
* Gives a visual idea of what sections are selected from selecting all
* Preserve keyboard navigation

Cons
* There are now 2 "All", which can be a bit weird. An alternative could be to replace "All" by "Entities" and "Social", but then we are back at throwing the term "Entities" at the face of new users (which nevertheless would have the good side of introducing the concept)

### DIFF
--- a/app/modules/general/scss/_top_bar.scss
+++ b/app/modules/general/scss/_top_bar.scss
@@ -1,4 +1,4 @@
-$search-section-width: 6em;
+$search-section-width-on-large-screen: 6em;
 
 nav#top-bar{
   // Use position:fixed to make the top-bar stick to the screen top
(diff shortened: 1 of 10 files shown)


jum-s (approved):

maxlath:
new proposition to address the keep all the search options available even on very small screens: wrap search settings when displaying search results

Screenshot_20201215-003102
Screenshot_20201215-003115

maxlath:
remerged as 1600d746

Issue #249, opened 2020-12-28
importer: no links provided to entities found


jum-s:
when a user imports new books (called candidates in the code), it would be nice to provide links to the entities found

![2020-12-28_12-29](https://user-images.githubusercontent.com/5363918/103213014-52a25800-4904-11eb-9448-82f99af4183a.png)

to be able to click on the tile, the authors and the isbn

maxlath:
fixed by #312

PR #250, opened 2020-12-28
importer: add link to found entities

jum-s:
solve #249

### DIFF
--- a/app/modules/inventory/views/add/candidate_row.js
+++ b/app/modules/inventory/views/add/candidate_row.js
@@ -36,7 +36,7 @@ export default Marionette.ItemView.extend({
     'click .remove': 'remov',
     // General click event: use stopPropagation to avoid triggering it
     // from other click event handlers
-    click: 'select'
+    'click input': 'select'
   },
 
   updateSelected (e) {
--- a/app/modules/inventory/views/add/candidate_row.js
+++ b/app/modules/inventory/views/add/candidate_row.js
@@ -19,7 +19,7 @@ export default Marionette.ItemView.extend({
   },
 
   initialize () {
-    this.editionModel = new Entity(this.model.attributes)
+    if (this.isSelectable()) { this.editionModel = new Entity(this.model.attributes) }
   },
 
   onShow () {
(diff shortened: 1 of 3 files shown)


maxlath:
* The problem with converting a candidate into an entity model (as done in `this.editionModel = new Entity(this.model.attributes)`) is that being an invalid or incomplete candidate is an accepted state (`isInvalid=true` or `needInfo=true`), while it's not for an entity, which seems to create different issues.
  * You can test an invalid candidate by passing an invalid ISBN http://localhost:3009/add/import?isbns=9780552167742
  * You can test an incomplete candidate by passing a valid but locally unknown ISBN http://localhost:3009/add/import?isbns=9782501151801 . I suspect that in this case, once info is entered by the user, you might need to handle the case where there is no entity associated with the candidate: the added info currently are only unstructured strings, that are later reconciled with the existing entities.
* Selected candidates have lost their green background, which was imo a helpful clue
* I would rather not make those links `target=_blank` as that breaks the general convention for internal links to open in the same tab: it should be fine to open in the same tab by default (can be done by just adding the class `showEntity` to the `a` elements), as long as hitting the previous button or coming back to the import page recovers the previous state. It's then still possible to open in a new tab by the standard means.
* I guess it's fine to not have the whole bar selectable, but a lot of users would probably enjoy having a larger click zone: having at least 1em around the checkbox clickable would be appreciable

jum-s:
I think i disagree working on the last point on this branch, it would be appreciable indeed but its an issue in itself as the client could have a common way to select things through checkboxes (ie. shelves, and other examples i cant think of now)
Will fix all the other points

jum-s:
tries to only convert a candidate into an entity model when necessary, but i could not find how to test it properly enough : 
- i could not incomplete candidate because dataseeds are powerful, 
- and passing an invalid candidate trigger the expected behavior => that is :

app/modules/inventory/views/add/candidate_row.js
```diff
@@ -17,13 +18,21 @@ export default Marionette.ItemView.extend({
     return base
   },
 
+  initialize () {
+    if (this.isSelectable()) { this.editionModel = new Entity(this.model.attributes) }
```
jum-s, line 22: is that better ? only actual selectable candidates are becoming entities

PR #251, opened 2020-12-29
inventory_network_nav: display friend requests and group invitations as sub sections

maxlath:
The primary motivation was to get invitations out of the groups list: now when declining an invitation, the list is updated, because it is bound to a more specific collection than just `app.groups`, which includes all groups you were you are member or invited when the session starts, and doesn't remove groups models from the collection when an invitation to join is declined. So I started looking at how to use a collection that would be updated, but once I got there, it seemed that there was an opportunity to make better users and groups lists, thus this proposition

## before
before

## after
after

## Ideas
* maybe those friend requests and invitations lists can get annoying, so it could make sense to have them wrapped up by default, with a counter, and wait for a click to unwrap

### DIFF
--- a/app/modules/inventory/scss/_inventory_nav_lists.scss
+++ b/app/modules/inventory/scss/_inventory_nav_lists.scss
@@ -1,6 +1,6 @@
 $nav-avatar-size: 5em;
 
-#usersList, #groupsList, #membersList{
+#usersList, #groupsList, #friendRequestsList, #groupsInvitationsList, #membersList{
   ul{
     @include display-flex(row, center, flex-start, wrap);
     margin: 0 auto;
@@ -48,7 +48,7 @@ $nav-avatar-size: 5em;
   }
 }
 
-#usersList, #groupsList{
+#usersList, #groupsList, #groupsInvitationsList{
   ul{
     @include radius;
     background-color: #f0f0f0;
@@ -97,5 +97,14 @@ $nav-avatar-size: 5em;
     background-color: $light-blue;
     background-color: $light-blue;
     color: $light-grey;
+  }
+}
+
+.friendRequestsWrapper, .groupsInvitationsWrapper{
+  padding: 0.5em;
+  margin-top: 0.5em;
+  @include radius;
+  &, ul{
+    background-color: darken(#eeeeee, 5%);
   }
 }
--- a/app/modules/inventory/views/templates/inventory_network_nav.hbs
+++ b/app/modules/inventory/views/templates/inventory_network_nav.hbs
@@ -10,6 +10,11 @@
       </div>
     </div>
     <div id="usersList"></div>
+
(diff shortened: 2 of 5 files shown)


app/modules/inventory/scss/_inventory_nav_lists.scss
```diff
@@ -99,3 +99,12 @@ $nav-avatar-size: 5em;
     color: $light-grey;
   }
 }
+
+.friendRequestsWrapper, .groupsInvitationsWrapper{
+  padding: 0.5em;
+  margin-top: 0.5em;
+  @include radius;
+  &, ul{
+    background-color: darken(#eeeeee, 5%);
```
jum-s, line 108: would be nice to use normalized colors

app/modules/inventory/views/templates/inventory_network_nav.hbs
```diff
@@ -23,5 +28,10 @@
       </div>
     </div>
     <div id="groupsList"></div>
+
+    <div class="groupsInvitationsWrapper hidden">
+      <p class="list-label">{{I18n 'invitations'}}</p>
```
jum-s, line 33: What about longer/more explicit text like : "You have been invited to those groups" or "groups invited to join"

app/modules/inventory/views/templates/inventory_network_nav.hbs
```diff
@@ -10,6 +10,11 @@
       </div>
     </div>
     <div id="usersList"></div>
+
+    <div class="friendRequestsWrapper hidden">
+      <p class="list-label">{{I18n 'friend requests'}}</p>
```
jum-s, line 15: What about longer/more explicit text like : "Users who wants to become friends with you" or "Users requesting friendship"

app/modules/inventory/views/templates/inventory_network_nav.hbs
maxlath: that would end up taking quite some space on small screens: a longer text could be set in `title` text maybe?

jum-s: sure, i was making this feedback because it did not occur to me straightfully that "invitations" meant "groups youre invited to" and it took me a some time getting the context (although i was fresh and caffeinated)

app/modules/inventory/views/templates/inventory_network_nav.hbs
jum-s: pending friend requests?

app/modules/inventory/views/templates/inventory_network_nav.hbs
jum-s: i think those "take some space on small screens" arguments only make sense for a defined language (english) but cannot be applied to all languages anyway

maxlath:
Closing as this PR code is now severly obsolete.

PR #252, opened 2020-12-30
Replace email tickets by a task

jum-s:
- open a `/tasks/works` route
- rely on `entitiesType` to not display any `scores` (which will be some work to extract if one day `collect-entities` for works is implemented), also to escape `relativeTask` as there is none possible (one suggestion for a suspect)


app/api/tasks.js
```diff
@@ -9,7 +9,9 @@ const byEntitiesUris = name => uris => {
 export default {
   byIds (ids) { return action('by-ids', { ids }) },
   byScore (limit, offset) { return action('by-score', { limit, offset }) },
+  byEntitiesType (params) { return action('by-entities-type', params) },
   bySuspectUris: byEntitiesUris('suspect'),
   bySuggestionUris: byEntitiesUris('suggestion'),
+  deduplicateWork: action('deduplicate-work'),
```
maxlath, line 15: update required after https://github.com/inventaire/inventaire/commit/b2a18f9
```suggestion
  deduplicateWorks: action('deduplicate-works'),
```

maxlath (changes requested):

app/modules/tasks/views/current_task.js
```diff
@@ -15,8 +15,13 @@ export default Marionette.LayoutView.extend({
   },
 
   onShow () {
-    this.showAuthor('suspect')
-    this.showAuthor('suggestion')
+    if (this.model.get('type') === 'deduplicate') {
```
maxlath, line 18:
```suggestion
    if (this.model.get('entitiesType') === 'humans') {
```

app/api/tasks.js
jum-s: reviews are precious :)

maxlath (approved):

PR #257, opened 2021-01-08
add an autocreate params to check on dataseed service

jum-s:
related to https://github.com/inventaire/inventaire/pull/470

### DIFF
--- a/app/api/entities.js
+++ b/app/api/entities.js
@@ -8,8 +8,9 @@ export default {
   // GET
   getByUris (uris, refresh, relatives) {
     uris = forceArray(uris).join('|')
+    const autocreate = true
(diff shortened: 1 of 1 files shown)


app/api/entities.js
```diff
@@ -8,8 +8,9 @@ export default {
   // GET
   getByUris (uris, refresh, relatives) {
```
maxlath, line 9: what about giving that control to consumers instead?
```suggestion
  getByUris ({ uris, refresh, relatives, autocreate }) {
```

maxlath (approved):

app/api/entities.js
jum-s: yagni ? well see when well need it no ?

PR #259, opened 2021-01-13
Go to next task when invalid

jum-s:
The idea is to have fluid experience when going from task to task.
Some task may be invalid, some may have already redirected suspect uri, in this case it simply go to the next one (but not when client come from a url with a task id)
Initially a commit which archive the task (by creating a state) was intended, but it doesnt seem necessary as it may disturb some tasks state. This client responsibility is only to ensure a smooth experience, not update tasks whatsoever.

### DIFF
--- a/app/modules/tasks/models/task.js
+++ b/app/modules/tasks/models/task.js
@@ -28,7 +28,9 @@ export default Backbone.Model.extend({
       const resolvedUri = model.get('uri')
       if (resolvedUri !== uri) {
         const context = { task: this.id, oldUri: uri, newUri: resolvedUri }
-        throw error_.new(`${name} uri is obsolete`, 500, context)
+        const err = error_.new(`${name} uri is obsolete`, 500, context)
+        err.code = 'obsolete_task'
+        throw err
       }
 
       return model.initAuthorWorks()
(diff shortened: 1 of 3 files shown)


app/modules/tasks/models/task.js
```diff
@@ -21,21 +21,24 @@ export default Backbone.Model.extend({
...
     .then(model => {
       const resolvedUri = model.get('uri')
       if (resolvedUri !== uri) {
-        const context = { task: this.id, oldUri: uri, newUri: resolvedUri }
-        throw error_.new(`${name} uri is obsolete`, 500, context)
+        if (isShownFromId) {
+          const context = { task: this.id, oldUri: uri, newUri: resolvedUri }
+          throw error_.new(`${name} uri is obsolete`, 500, context)
+        }
+        return this.showNextTask()
```
maxlath, line 34: This is view logic (as hinted by the `show` terminology) and shouldn't be in the model. It might even have been the initial intention has `this.showNextTask` is not defined on the model but on `tasks_layout`, thus the error `this.showNextTask is not a function`

To implement this behavior from the view, I would suggest to catch the error thrown by the model from the view, and call `this.showNextTask` if the error matches that known exception

maxlath (changes requested):

app/modules/tasks/models/task.js
jum-s: definetly

maxlath (approved):

PR #264, opened 2021-02-07
add data serialization missing to display suggestions

jum-s:
maybe there is a reason why this was unfinished.
In case there is no reason, here is the very last step to display suggestions
#263

### DIFF
--- a/app/modules/entities/views/publisher_layout.js
+++ b/app/modules/entities/views/publisher_layout.js
@@ -20,6 +20,7 @@ export default TypedEntityLayout.extend({
 
   initialize () {
     this.model.initPublisherPublications()
+    this.displayMergeSuggestions = app.user.hasAdminAccess
   },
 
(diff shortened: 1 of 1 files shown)


app/modules/entities/views/publisher_layout.js
```diff
@@ -20,6 +20,13 @@ export default TypedEntityLayout.extend({
 
   initialize () {
     this.model.initPublisherPublications()
+    this.displayMergeSuggestions = app.user.hasAdminAccess
+  },
+
+  serializeData () {
+    return _.extend(this.model.toJSON(), {
+      displayMergeSuggestions: this.displayMergeSuggestions
+    })
```
maxlath, line 29: this method definition is not needed, as the view inherits `serializeData` from `TypedEntityLayout`

maxlath (changes requested):

PR #272, opened 2021-02-20
add identifiers to authors

jum-s:
the current state of this pr is intended to be a base of discussion to start integrating some identifier to authors layout
with the idea to improve tasks suggestions (as described in #501)
maybe we could wrap the identifiers to only display it on a user's click, maybe we could only display it in a taks context, maybe its fine like this now, i dont have a strong opinion on any of those suggestions

### DIFF
--- a/app/lib/handlebars_helpers/identifiers.js
+++ b/app/lib/handlebars_helpers/identifiers.js
@@ -1,6 +1,14 @@
 export default {
-  'wdt:P496': {
-    text: _.identity,
-    url (id) { return `https://orcid.org/${id}` }
-  },
+  'wdt:P213': id => `https://isni.oclc.org/xslt/DB=1.2//CMD?ACT=SRCH&IKT=8006&TRM=ISN%3A${id}`,
+  'wdt:P214': id => `https://viaf.org/viaf/${id}`,
+  'wdt:P227': id => `https://d-nb.info/gnd/${id}`,
+  'wdt:P244': id => `https://id.loc.gov/authorities/${id}`,
+  'wdt:P268': id => `https://catalogue.bnf.fr/ark:/12148/cb${id}`,
+  'wdt:P269': id => `https://www.idref.fr/${id}`,
+  'wdt:P271': id => `https://ci.nii.ac.jp/author/${id}`,
+  'wdt:P496': id => `https://orcid.org/${id}`,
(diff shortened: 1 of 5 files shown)


app/lib/handlebars_helpers/identifiers.js
```diff
@@ -0,0 +1,14 @@
+export default {
+  'wdt:P213': id => { return `https://isni.oclc.org/xslt/DB=1.2//CMD?ACT=SRCH&IKT=8006&TRM=ISN%3A${id}` },
```
maxlath, line 2: syntactic preference would be to use the arrow function implict return for the beauty of that simpler one-liner
```suggestion
  'wdt:P213': id => `https://isni.oclc.org/xslt/DB=1.2//CMD?ACT=SRCH&IKT=8006&TRM=ISN%3A${id}`,
```

maxlath:
I'm all in favour of making those ids and URLs available to people who would appreciate it, but there are a lot of external ids and it will end up being extremely cluttered for most uses: I was thinking we could make this choice possible in user preferences; the user could check/uncheck all the properties to use to fit there interest. I would, for instance, probably keep the BNF activated, but disable instragram, facebook, and most libraries and databases I don't really use and could still easily access via Wikidata, without paying the prices of a page full of ids

app/lib/handlebars_helpers/identifiers.js
jum-s: yeah i know, but somehow webpack doesnt like it and fails to compile...

maxlath: what's webpack error message?

jum-s:
Would you consider a entities tab in the settings. Somewhere around : https://github.com/inventaire/inventaire-client/blob/master/app/modules/settings/views/settings_layout.js#L40

maxlath:
That could be on the display settings page, in a subsection named "entities" or some category name that could makes more sense to new comers

jum-s:
since display setting page is not yet on master, i would wait for #205 to be merged

app/lib/handlebars_helpers/identifiers.js
jum-s: done 884c4b4

jum-s:
done with #339

PR #273, opened 2021-02-26
remove the listing dropdown option "remove item"

jum-s:
for 2 reasons:
  - a remove item action button is already available on the item modal
  - this dropdown is about listing/visibility which has little (not to
  say nothing) to do with the rest of the options

only the template is affected since the layout is still triggered by the
remove class of the html element in the modal

related PR: #271 

message to reviewer: feel free to close it if you disagree

### DIFF
--- a/app/modules/inventory/views/item_card.js
+++ b/app/modules/inventory/views/item_card.js
@@ -36,7 +36,6 @@ export default ItemItemView.extend({
   events: {
     'click a.transaction': 'updateTransaction',
     'click a.listing': 'updateListing',
-    'click a.remove': 'itemDestroy',
     'click a.itemShow': 'itemShow',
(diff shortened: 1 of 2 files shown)


maxlath:
I agree that it's redundant on the `item_show` layout, but it can be quite useful to have it on the `item_card` layout, which consumes that same `item_visibility_box` template: what do you think of adding a flag to display it in one case and not the other?

jum-s:
I dont think its coherent to click on a visibility icon and to be offered to delete the item.
I wont be shocked if i have to first display the item model in order to delete an item.
An item visibility and the possibility of deleting it are very different actions, imo, they shall not end up within the same dropdown.

If those arguments dont convince you, i will make the split.

PR #274, opened 2021-02-26
deduplicate layouts rewrites with Svelte

maxlath:
The future is here!

That was quite adventurous to start with those complex layouts, but it allows to experiment with some of the benefits of moving to Svelte. The biggest problem was that rewriting functions Backbone-free creates some duplicated code, as long as we can't totally get rid of the Backbone.Marionnette layouts still relying on those legacy functions. Given this, I would be in a favor of not being too zealous on duplicated code during the transition, and generally allow for experimentations as we still need to figure-out the new best practices: it's safe to say that our future self will not a lot about Svelte and be able to easily cleanup our newbie mistakes (as long as we don't go into too intricated hacks of course)

The general feeling is very positive: it was somewhat painful to get things started, but the joy of different things that use to be pain points becoming trivial makes it totally worth it [insert lyrical ode to Svelte]. I tried to roughly stay on a feature parity rewrite, but some features were added along the way as it was too tempting :D


maxlath:
Conventions adopted in concertation with @jum-s:
* We should let libs used by Backbone views alone, as they are full of Backbone Models and Collections, and recreate functions, with possibly duplicated code, in separated files, typically on the pattern `author.js` (legacy Backbone function) / `author_alt.js` (vanilla functions to serve the needs of Svelte components), until the Backbone views are made obsolete, their dependencies deleted, and the `*_alt.js` libs all renamed `*.js`
* Components CSS should stay in Svelte modules (instead of being imported from a separate file) to benefit from Svelte capacity to report obsolete selectors/rules

app/modules/entities/components/deduplicate_authors_names.svelte
```diff
@@ -0,0 +1,55 @@
...
+  import preq from 'lib/preq'
+  import Spinner from 'modules/general/components/spinner.svelte'
+  import { createEventDispatcher } from 'svelte'
+
+  const waitingForNames = fetchNames()
+  let names
+
+  async function fetchNames () {
+    const res = await preq.get(app.API.entities.duplicates)
+    names = res.names
```
jum-s, line 11: is `fetch` the new `grab` ? shall we stick to this verb for anything related to a API query ?

app/modules/entities/components/deduplicate_works.svelte
```diff
@@ -0,0 +1,260 @@
...
+
+  export let author
+
+  let wdWorks = []
+  let invWorks = []
+  let index = -1
+  let merging = false
+  let wdDisplayLimit = 10
+  let invDisplayLimit = 10
+  let candidates, allWorksByPrefix, allCandidateWorksByPrefix, error, filterPattern, displayedWdWorks, displayedInvWorks, windowScrollY, wdBottomEl, invBottomEl
```
jum-s, line 18: what about making explicit only collections variables, aka are they either array or object ? 
example with random type attribution:

```js
let candidates = [], wdWorks = []
let allWorksByPrefix = {}, allCandidateWorksByPrefix = {}, displayedWdWorks = {}, displayedInvWorks = {}
let error, filterPattern, windowScrollY
```

jum-s (approved):

PR #285, opened 2021-06-03
Svelte settings rewrites

jum-s:
with the parity feature in mind, this propose a refresh of settings layout.
Its pretty light in the end since its heavily reuse of the existing modals (for profile pictures and location updates). But considering svelte is quite a new framework to handle, i would postpone the modal extraction from backbone for the next PR.

a quick interrogation : 
- this PR propose to give the responsibility of toggling periodicity to `notifications.svelte`. BUT maybe, this is actually the responsibility of `Toggler.svelte`, which could be passed a general action (ie. `{ action: togglePeriodicity, actionTrigger; "unselect" }`). Something similar to the modal behavior today

pre-merge action: merge i18n PR [#4](https://github.com/inventaire/inventaire-i18n/pull/4)


maxlath (changes requested):
This is a purely functional review as I haven't looked at the code yet:

## Regressions
* when changing language, the page takes a moment to reload without any visual feedback indicating that we have to wait
* presentation textarea doesn't autosize
* broken email availability check: somehow, whatever the email, requests are made with `email=filter` (`/api/auth?action=email-availability&email=filter`)
* currently in prod, when all notifications are disabled (via the "Global > receive email notifications" setting), its not possible to change finer notifications settings as they have no effect: this behaviour has been broken, as one can now disable email notifications globally and still change finer settings, which is confusing

## Comments
* I don't think the "Download your profile" section brings much, as that profile data is already available from the "Data" section
* The profile data file shouldn't be named `profile.csv` as it's a JSON file, and "All information stored about you" is a rather exaggerated statement as it only returns the user document
* The "Check the documentation" link doesn't look like a link (it's already the case in prod, but could be fixed)
* "forgot password" now looks like primary text (="everybody is gonna read it"), rather than secondary text (="people looking for it should find it")

app/lib/components/update_password.svelte
```diff
@@ -0,0 +1,70 @@
...
+      if (err.statusCode === 401) return flashCurrentPwdErr('wrong password')
+    }
+    try {
+      await app.request('password:update', currentPassword, newPassword)
+      showFlashNewPassword({
+        priority: 'success',
+        message: I18n('done')
+      })
+    } catch {
+      return flashCurrentPwdErr('something went wrong, try again later')
```
maxlath, line 34: *comment true for all occurrences of `catch {`*

if you ignore the error, we will never know what happened (the error won't be reported to the server): if you prefer to not show the error message to the user, at least log it
```suggestion
    } catch (err) {
      // Logs the error and report it
      log_.error(err)
      return flashCurrentPwdErr('something went wrong, try again later')
```

maxlath (changes requested):

app/lib/components/update_password.svelte
jum-s: right, its kind of a new feature as there was no catch originally.

> if you prefer

i dont really know what i prefer tho, any suggestion ?

jum-s:
every regressions should be fixed
all comment have been taken into account

maxlath (changes requested):
Remaining issues:

### Location
* [x] "no position set" still displayed after a position was set (and the old position being displayed when the position is removed)

### Email
* [x] "This email is valid and available." should probably not be displayed as it could be confused with the "done" message, and the user might then not realize they didn't save
* [x] when saving, "undefined" is displayed rather than "waiting"
* [x] the `new_confirmation_email` is displayed as text rather than interpreted as html, resulting in `<br>`  being displayed as text

### Password
* [x] missing a "waiting" spinner when saving

### Notifications
* [x] toggling global notifications doesn't seem to actually save the setting on the server (see [video](https://tmpshared.maxlath.eu/global_notifications_bug.mp4
)), other notification settings seem fine

### Data
* [x] when clicking user profile "Download CSV" button, the profile is actually returned as JSON 
~* [ ] nothing happens when clicking inventory "Download CSV" button~
~* [ ] nothing happens when clicking user profile "Download JSON" button~

jum-s:
download inventory CSV and download profile json works on my setup, doesnt work even if waiting a bit on your side ? does the server reply a 200 ?
 "This email is valid and available." has been changed to an error if email is not available

still todo : "no position set" still displayed after a position was set (and the old position being displayed when the position is removed)

maxlath:
sorry, "download CSV" and "download JSON" issues were due to bad local data (items with deleted entities) making the server respond with 500: my bad

maxlath (approved):

app/lib/components/password_input.svelte
```diff
@@ -0,0 +1,46 @@
...
+  }
+</script>
+
+<!--
+Prefer on:change to bind:value since svelte cannot dynamically change a value if another one is binded.
+aka `type={pwdInputType} bind:value={password}` is forbidden since generated code is different for different kinds of input. -->
+<input type={pwdInputType} on:change="{e => changePwd(e.target.value) }" title={name}>
+<Flash bind:state={flash}/>
+<div class="showPasswordWrapper">
+  <input type="checkbox" class="showPassword" id="show{name}" bind:checked={showPassword} on:click={togglePassword}>
```
maxlath, line 29: Integration with browser password managers was lost: that regression seems to pre-date this PR, but it would be cool to recover it while we are here

maxlath (changes requested):

app/lib/components/password_input.svelte
maxlath: addressed in c27dbca

maxlath (approved):

maxlath:
@jum-s the commits I pushed introduce the use of a global `$user` store, on par with the globla `app.user` Backbone model. This allowed to simplify the settings component and should ease future rewrites depending on the user data (ex: top bar or views displaying the username or the profile picture)

PR #287, opened 2021-06-30
Rewrite activity to contributions svelte

jum-s:
rename activity -> contributions, to avoid mixing with activity streams

depends of server PR [539](https://github.com/inventaire/inventaire/pull/539)

### DIFF
--- a/app/modules/entities/components/contributions.svelte
+++ b/app/modules/entities/components/contributions.svelte
@@ -1,0 +1,15 @@
+<script>
+  import ContributionsPeriod from './contributions_period.svelte'
+</script>
+<div id="activityLayout">
+  <ContributionsPeriod title='last day' period=1/>
+  <ContributionsPeriod title='last week' period=7/>
+  <ContributionsPeriod title='global'/>
+</div>
+<style>
+  #activityLayout{
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+  }
+</style>
--- a/app/modules/entities/components/contributions_period.svelte
+++ b/app/modules/entities/components/contributions_period.svelte
@@ -8,12 +8,12 @@
   export let title, period
 
   const getContributionsData = async () => {
-    const res = await preq.get(app.API.entities.activity(period))
+    const res = await preq.get(app.API.entities.usersContributions(period))
     usersData = await addUsersData(res)
   }
 
   const addUsersData = async res => {
-    let { activity: contributionRows } = res
+    let { contributions: contributionRows } = res
     if (contributionRows.length === 0) return []
     contributionRows = contributionRows.slice(0, 10)
     const usersIds = contributionRows.map(_.property('user'))
--- a/app/modules/entities/components/contributions_period.svelte
+++ b/app/modules/entities/components/contributions_period.svelte
@@ -8,7 +8,7 @@
   export let title, period
 
   const getContributionsData = async () => {
-    const res = await preq.get(app.API.entities.usersContributions(period))
+    const res = await preq.get(app.API.entities.usersContributionsCount(period))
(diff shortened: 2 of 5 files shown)


app/modules/entities/components/contributions.svelte
```diff
@@ -0,0 +1,16 @@
+<script>
+  import ContributionsPeriod from './contributions_period.svelte'
+</script>
+<div id="activityLayout">
```
maxlath, line 4: Could we keep as a convention to add one empty line as padding between script/template/style blocks? That improves readability imo

app/modules/entities/components/contributions.svelte
```diff
@@ -0,0 +1,16 @@
...
+  <ContributionsPeriod title='last day' period=1/>
+  <ContributionsPeriod title='last week' period=7/>
+  <ContributionsPeriod title='global'/>
+</div>
+<style lang="scss">
+  @import 'app/modules/general/scss/utils';
```
maxlath, line 10:
```suggestion

```
no scss or utils are used here

app/modules/entities/components/contributions_period.svelte
```diff
@@ -0,0 +1,102 @@
...
+      <div class="cell user {userData.user.deleted ? 'deleted' : ''}">
+        <a href="/u/{userData.user._id}/contributions" class="link" on:click="{e => showUserContributions(e, userData.user)}">{userData.user.username}</a>
+      </div>
+      <div class="histogram">
+        <div class="contributions">
+          {userData.contributions}
+        </div>
+        <div class="bar" style="width: {userData.contributions * 100.0 / highest}px;"></div>
+      </div>
+    </div>
```
maxlath, line 54: missing indentation within the each block

app/modules/tasks/views/tasks_layout.js
```diff
@@ -117,10 +117,8 @@ export default Marionette.LayoutView.extend({
         collection: model.suspect.mergeSuggestions,
         currentTaskModel: model
       })
-      if (model.suspect) {
-        this.relativeTasks.show(newRelativeTasks)
-        return this.updateRelativesCount(model)
-      }
+      this.relativeTasks.show(newRelativeTasks)
+      return this.updateRelativesCount(model)
```
maxlath, line 121: how is that connected to this PR?

app/modules/entities/components/contributions.svelte
maxlath: Also, `activityLayout` -> `contributonsLayout`?

maxlath (changes requested):
minor details

Also, removing obsolete files such as `app/modules/entities/views/activity_layout.js` could be part of this PR

app/modules/entities/components/contributions.svelte
jum-s: yeah not really, the more it goes the more empty lines only make files longer to scroll, i would actually be for deleting all empty lines of the project :)

maxlath: many empty lines are the sign that a function is too big and should be split in sub functions, and eventually in different module files (yeah modularization!), but in the case of Svelte components, we can't just split those 3 distinct roles over 3 files, they have to live in the same file, while serving different purposes, so separating by an empty line seems the decent thing to do (and [what people seem to be doing](https://github.com/search?l=Svelte&q=style&type=Code))

jum-s: those text blocks are distinctive enough to me, from js to html to css, within the first second i can see which is which, therefore an extra line to distinct within the same 'language' makes sense but not here, it only makes you scroll more

PR #289, opened 2021-08-03
Broken wiki link in credits

AutomCoding:
I also believe the anchor text should be "statutes" in English, but don't know where to change that


maxlath:
indeed, thanks!

PR #290, opened 2021-09-08
add a fedeversable setting checkbox

jum-s:
now that [`activity-pub`](https://github.com/inventaire/inventaire/pull/538) server branch is merged, allow user to become fediversable

### DIFF
--- a/app/lib/components/notification_toggler.svelte
+++ b/app/lib/components/notification_toggler.svelte
@@ -36,8 +36,9 @@
   @import 'app/modules/general/scss/utils';
   input{
     margin: 0;
-    /*otherwise chrome do not display any checkbox*/
-    appearance: auto;
+  }
+  label {
+    font-size: 1rem
   }
   .wrapper{
     display: flex;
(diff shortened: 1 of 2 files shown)


maxlath:
the `input { appearance: auto; }` makes that checkbox appear as a vertical bar:
b
Removing that rule here and in `app/lib/components/notification_toggler.svelte` doesn't seem to break in Chrome, contrary to what the comment says: can we remove those rules?
Also, a11y point, those checkbox labels are too small, what about adding `label{ font-size: 1rem }` here and in `app/lib/components/notification_toggler.svelte`?

maxlath (changes requested):
the `input { appearance: auto; }` makes that checkbox appear as a vertical bar:
b
Removing that rule here and in `app/lib/components/notification_toggler.svelte` doesn't seem to break in Chrome, contrary to what the comment says: can we remove those rules?
Also, a11y point, those checkbox labels are too small, what about adding `label{ font-size: 1rem }` here and in `app/lib/components/notification_toggler.svelte`?

jum-s:
done in `e925328f`

maxlath (approved):

PR #291, opened 2021-09-15
experimentation on how to reuse entities data from another instance

maxlath:
client part of https://github.com/inventaire/inventaire/pull/548

When the server sends a `remoteEntities` URL in the client config ([`/api/config`](https://inventaire.io/api/config))
* Fetches entities data from that remote instance (made possible by the open CORS policy https://github.com/inventaire/inventaire/pull/543)
* Uses entities images from that remote instance
* Searches entities on that remote instance
* Disable `admin` and `dataadmin` views (as those rights apply to the local entities, but can't be assumed for the remote instance)

### DIFF
--- a/app/api/entities.js
+++ b/app/api/entities.js
@@ -4,17 +4,26 @@ const { action } = endpoint('entities')
 
 const CustomQuery = actionName => (uri, refresh) => action(actionName, { uri, refresh })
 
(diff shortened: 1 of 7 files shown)


maxlath:
Closing as this PR code is now severely obsolete, but the idea could be reimplemented when we come back to that subject.

Issue #292, opened 2021-10-08
Accessibility issues


jum-s:
This issue was produced thanks to the accessibility audit done by the [Accessibility Foundation](https://www.accessibility.nl/english). The list below is not comprehensive and the focus should not be limited to this list.
Any of those point in the list can be tackled independently (please add commit id when fixing a point)
Some previous accessibility work has been done with #248

page specific:
- [x] items: add an alt text on items cover (1.1.1)
- [x] welcome page: add an alt text on recent items cover (1.1.1)
- [x] settings/notifications: link the question through fieldset and legend elements (1.3.1)
- [x] settings/account : "Danger Zone" should be a darker red (1.4.3)
- [x] network: invite a friend: input field placeholder "Email address" should be darker (1.4.3)
- [x] items cascade: send request background should be darker (1.4.3)
- [x] item: title "layout-type-label" should be darker (1.4.3)
- [x] login: at 200% zoom, login buttons should be displayed (1.4.4)
- [x] login: buttons "Show", "Create an account" and "Forgot your password?" should be  keyboard operable (2.1.1)
- [x] login: add some hidden text using CSS that describes the "x" button to close the error message. (4.1.2)
- [x] item pop-up: add some hidden text using CSS that describes the "x" button to close pop-up. (4.1.2)
- [x] settings/notifications: error notifications, closing button must have a name (4.1.2)
- [x] entity/work: add name, role and value to toggle button (4.1.2)
- [x] welcome: add to "Our mission & values" button an aria-expanded="true" attribute (4.1.2)
- [x] search bar: add aria attributes to search field (3.2.2)

general:
- [x] add autocomplete attribute to inputs (1.3.5)
- [x] give a second hint to keyboard focus (ie. border or underline element) (example: [mastodon join page input fields](https://mamot.fr/about)) (1.4.1)
- [x] at 200% zoom, language selection menu should be displayed (1.4.4)
- [x] input fields borders should have a higher contrast (1.4.11)
- [x] languages drop-down menu should be keyboard accessible (2.1.1)
- [x] search: without pressing 'Escape', results menu should close when tabbing (2.4.3)
- [ ] improve visible focus in general (2.4.7)
- [x] <html> should have the correct lang attribute (3.1.1)
- [x] give foreign-language texts a language identification (3.1.2)
- [x] empty feedback should not be sent (3.3.1)
- [ ] search: give a label with instructions what the input field is for (placeholder's not enough) (3.3.2)
[...]

PR #293, opened 2021-10-13
addressing a11y audit

maxlath:
addressing #292


maxlath:
@jum-s yes, deleting ids if unused by css or js seems right, as those id would otherwise not serve any purpose and just clutter up the global id namespace. Fun fact: elements with ids are accessible as global variables at `window[id]`: every time an id is set, a global variable is created. It doesn't change anything security-wise as that element was accessible from the DOM anyway, but it does sound bad ^^

jum-s (approved):

jum-s:
rebased, ready to merge

PR #307, opened 2022-01-07
Recover email validation request

maxlath:
It use to be possible to request a new email with a token to validation the user email address; that was lost when the settings got rewritten in Svelte, and this PR brings that back

### DIFF
--- a/app/modules/settings/components/email_validation.svelte
+++ b/app/modules/settings/components/email_validation.svelte
@@ -1,0 +1,68 @@
+<script>
+  import { icon } from 'lib/utils'
+  import { I18n } from 'modules/user/lib/i18n'
+
+  let validationEmailRequested = false
+
+  async function emailConfirmationRequest () {
+    validationEmailRequested = true
+    try {
+      await app.request('email:confirmation:request')
+    } catch (err) {
+      validationEmailRequested = false
+      throw err
+    }
+  }
+</script>
+
+<div class="emailValidation">
+  {#if validationEmailRequested}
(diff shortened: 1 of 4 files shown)


jum-s (changes requested):
this is graphically getting a lot of user's attention, but its neither an error nor even a problem, only optional button if one havent received an email.

Instead i would integrate it the same way as other points in the settings
Title: "unverified email"
Message: in warning color with the current icon
Action button: like any other

on the blocking points:
-    align text on the left (as every text of the whole settings)
-    keep action button to a regular color, this black is not very spread (i would wait to refactor modals with svelte to introduce new action colors)
-    no background color for this section (same as any other section in settings)

maxlath:
@jum-s style harmonization done in 8f548eb

app/modules/settings/components/email_validation.svelte
```diff
@@ -15,54 +15,17 @@
...
+  <p>
+    {@html icon('check')}
+    {@html I18n('new_confirmation_email')}
+  </p>
+{:else}
+  <p>
+    {@html icon('warning')}
+    {@html I18n('email_wasnt_verified')}
+  </p>
+  <button class="button light-blue radius" on:click={emailConfirmationRequest}>
```
jum-s, line 28:
```suggestion
  <button class="light-blue-button radius" on:click={emailConfirmationRequest}>
```

maxlath: done in 59c863a

jum-s (approved):

PR #309, opened 2022-02-16
allow to search for wd uri with lowercased id

jum-s:


app/modules/search/lib/find_uri.js
```diff
@@ -1,11 +1,14 @@
...
-import wdk from 'lib/wikidata-sdk'
+import { isInvEntityId } from 'lib/boolean_tests'
 import { looksLikeAnIsbn, normalizeIsbn } from 'lib/isbn'
 
 export default function (text) {
   text = text.trim()
-  if (isEntityUri(text)) return text
-  if (wdk.isWikidataItemId(text)) return 'wd:' + text
+  if (isWikidataId.test(text)) text = 'wd:' + text
+  if (caseInsensitiveEntityUri.test(text)) return text.replace(/wd:q/, 'wd:Q')
```
maxlath, line 7: [non-blocking]
no need for a RegEx here
```suggestion
  if (caseInsensitiveEntityUri.test(text)) return text.replace('wd:q', 'wd:Q')
```
unless you want to precise that the replaced pattern must be at the beginning
```suggestion
  if (caseInsensitiveEntityUri.test(text)) return text.replace(/^wd:q/, 'wd:Q')
```
(but that's made unnecessary by the `caseInsensitiveEntityUri.test` which wouldn't let pass a string with 'wd:q' elsewhere than at the beginning)

PR #310, opened 2022-02-16
Use NodeJS standard `imports` to replace module aliases

maxlath:
Currently, we have path aliases defined as `alias` in package.json, consumed by webpack [`resolve.cjs`](https://github.com/inventaire/inventaire-client/blob/029946e/bundle/resolve.cjs) and by mocha via the [`custom-loader`](https://github.com/inventaire/inventaire-client/blob/fec0347/custom-loader.js). This setup broke when upgrading to NodeJS v16 as it was [relying on experimental node flags](https://github.com/inventaire/inventaire-client/blob/fec0347/scripts/run_unit_tests).

This PR suggests to fix this by moving to the [NodeJS `imports`](https://nodejs.org/docs/latest-v16.x/api/packages.html#subpath-imports) syntax for all JS assets, and then let aliases for other types of assets bundled by webpack, but not by tests. 

A note on the `.js` extension: webpack can resolve JS files without having to specify the `.js`, but from node's point of view ["the file extension is always necessary"](https://nodejs.org/docs/latest/api/esm.html#terminology), so those extensions are required for other consumers than webpack, namely mocha, unless using an `imports` alias that includes `.js`


package.json
```diff
@@ -9,32 +9,29 @@
...
-    "api": "./app/api",
-    "assets": "./app/assets",
-    "behaviors": "./app/modules/general/behaviors",
-    "lib": "./app/lib",
-    "modules": "./app/modules",
-    "vendor": "./vendor"
+  "imports": {
+    "#app/*": "./app/*.js",
+    "#assets/*": "./app/assets/*.js",
+    "#behaviors/*": "./app/modules/general/behaviors/*.js",
```
jum-s: since we are here, update svelte introduction

```suggestion
    "#behaviors/*": "./app/modules/general/behaviors/*.js",
    "#components/*": "./app/modules/general/components/*.js",
```

troubleshooting.md
```diff
@@ -139,4 +139,4 @@ Changing npm version or running `npm rebuild node-sass` should solve the issue.
 
 ## Svelte
 ### No reactivity despite an assignment in a nested object
-Assignments in nested objects won't trigger a refresh, until the top level object is reassigned. See https://github.com/sveltejs/svelte/issues/3212#issuecomment-510263274
+Assignments in nested objects won't trigger a refresh, until the top level object is reassigned. See https://github.com/sveltejs/qsvelte/issues/3212#issuecomment-510263274
```
jum-s: typo

jsconfig.json
```diff
@@ -3,13 +3,12 @@
   "compilerOptions": {
     "baseUrl": "./app",
     "paths": {
-      "api/*": [ "api/*" ],
-      "assets/*": [ "assets/*" ],
-      "behaviors/*": [ "modules/general/behaviors/*" ],
-      "lib/*": [ "lib/*" ],
-      "modules/*": [ "modules/*" ],
+      "#assets/*": [ "assets/*" ],
+      "#behaviors/*": [ "modules/general/behaviors/*" ],
```
jum-s:
```suggestion
      "#behaviors/*": [ "modules/general/behaviors/*"
      "#components/*": [ "modules/general/components/*" ],
```

troubleshooting.md
maxlath: done in 64e44fb2d

jsconfig.json
maxlath: with 2fcbf9bd4 jsconfig.json is a derived from package.json

PR #311, opened 2022-02-17
[WIP] Entity editor svelte rewrite

maxlath:


maxlath:
This rewrite largely sticks to the previous implementation, with a few notable exceptions:
- the entity_create component is not using the entity_edit component to make the differences more readable, and not overload entity_edit with `if createMode` forks
- multi_entity_edit was removed in favor of a much more specific entity_create_edition_and_work_from_isbn, has that was the only multi_entity_edit use-case, and it allows to make the process hopefully much easier to understand
- the entity_edit admin section has been replaced by:
  - a entity_edit_menu, gathering the different actions available to a user (depending on their access level)
  - a merge layout with entity autocomplete (which was the occasion to decouple the entity_autocomplete_selector from entity_edit, allowing to reuse it in other components, such as the importer)

jum-s:
as for starters (more to come)

feedback UX:
 - not having a save button (which is not feature parity), makes it unusable for people without keyboard. Even with a keyboard, not every people have the reflex to hit enter to validate.

 autocomplete
 - inside autocomplete, "close" button could say what it close? suggestion: "close suggestions"
 - when i press arrow down to see the suggestions, when i reach the last suggestion and i hit down arrow again, focus is back on text input, while i feel it could go back to first suggestion (at least it shall not unfocus autocomplete)

 image
 - reset button icon is confusing since its the same round shape as the two other icons but action is totally different (suggestion: either use fast-forward icon or display "reset" text next to icon)
 - when inserting a URL in "from a URL" input, no information on screen is displayed (in console only), if image couldnt have been converted. What about a flash message in `image_value_input.svelte` on error ? because no i have no idea if image has been uploaded or any status

datepicker
- not feature parity: going from January to February (aka from 1 to 2) in month column with a keyboard, i need to press arrow key up (and not arrow key down as intuitively scrolling down to find the month)

bugs:
- i can click on ISBN and isbn-13 claims while they shall not be edited
- when i open an edition editor, i click on "collection: add", then i type the collection name,  then i create a new inv collection by clicking on "Create a new Collection: foo", then "foo" is not displayed in the editor (but collection is created successfully)
- when i click on "Add" on a link claim, ie. an empty official website, then i click on the cross to close claim edition, i have an error "URL is null" (stack trace: `app/lib/components/link.svelte:7`)
- when i click on editing an exiting P1476 edition title: i have error`with a flash `TypeError: expected string, got undefined (undefined)`

jum-s:
2nd round

- when i want to add a genre to a work, then i type the letter "e", then a work is suggested, should be a genre
- when i click on the delete button of an existing "subject" P921, then i click on "cancel", then i press Enter, nothing is done (i expected the entity previously remove to be validated again)
- when i click on the delete button of an existing "subject" P921, i would like to add another claim value but no "add" button is shown to me
- when i add an author to a work, i do something to flash an error, then i add another author, the flash displayed is not disappearing (aka not reset to null)

small screen
- move add button to the right by default [non-blocking]
- actions menu is not visible (looks like its off screen on the left) quick hack suggestion: move actions button to the right part of the screen, dropdown should follow.
- display edit-mode-buttons as column to take less space (feature parity) (would need to remove absolute position of class `autocomplete` element) 

actions menu:
- without having connected user account to wikidata, when i click on "move to wikidata", there is a console error `this.model is undefined` stacktrace: `wikidata_edit_intro.js:11`
- renaming suggestion "open entity history"=>"entity history" (feature parity) open as the advantage of providing an the action verb, but open what ? other verb suggestion "see" or "view"

maxlath:
addressing the 2nd round of comments:

> when i want to add a genre to a work, then i type the letter "e", then a work is suggested, should be a genre

can't reproduce, typing "e" does trigger the correct search request: http://localhost:3006/api/search?types=genres&search=e&lang=fr&limit=10&offset=0&exact=false

> when i click on the delete button of an existing "subject" P921, then i click on "cancel", then i press Enter, nothing is done (i expected the entity previously remove to be validated again)

would closing the edit mode to show that the original value was restored work? I tried that in e35847355

> when i click on the delete button of an existing "subject" P921, i would like to add another claim value but no "add" button is shown to me

should have been fixed by 35d0d8089

> when i add an author to a work, i do something to flash an error, then i add another author, the flash displayed is not disappearing (aka not reset to null)

fixed by 56d3de53e

#### small screen
> move add button to the right by default [non-blocking]

it seems fine on the left, imo, could you do a screenshot of the configuration in which it seems problematic?

> actions menu is not visible (looks like its off screen on the left) quick hack suggestion: move actions button to the right part of the screen, dropdown should follow.

can't reproduce, at which screen size does the actions menu disappear?

> display edit-mode-buttons as column to take less space (feature parity) (would need to remove absolute position of class autocomplete element)

If you mean having the button go below the input, then it's addressed by a2c50b9
If you mean having edit-mode-buttons themselves displayed on a column, that doesn't seem to be the case on the pre-rewrite version either, see screenshot:
edit-mode-buttons

####  actions menu:
> without having connected user account to wikidata, when i click on "move to wikidata", there is a console error this.model is undefined stacktrace: wikidata_edit_intro.js:11

fixed by fb69eb1

> renaming suggestion "open entity history"=>"entity history" (feature parity) open as the advantage of providing an the action verb, but open what ? other verb suggestion "see" or "view"

done in b25f47b

jum-s:
non commented comments could be considered discarded.

First round 

>> when i press arrow down to see the suggestions, when i reach the last suggestion and i hit down arrow again, focus is back on text input, while i feel it could go back to first suggestion (at least it shall not unfocus autocomplete)

>Contrary to hitting tab, the focus never leaves the text input when hitting up or down (as you can see by setting `:focus{ background-color: red; }`)

and I would also keep it that way...

>This allows to keep the possibility to correct the text input, so I would rather keep it that way. There is indeed an issue with this up/down navigation though, when reaching the end of the list, fixed by 53c95a7e1

...when im hitting one more down after the last suggestion, it could focus on the text input (that's fine), but suggestion focus could come back to the first suggestion. Use case i met before: maybe i missed the right one, so i go back to the top suggestions to check.
its non-blocking if its more complicated than a `highlightedIndex = 0` to add somewhere.


Second round 

>> actions menu is not visible (looks like its off screen on the left) quick hack suggestion: move actions button to the right part of the screen, dropdown should follow.

>can't reproduce, at which screen size does the actions menu disappear?



it doesnt disappear, its not fully visible, looks like `align-right` is true on small screens. As we are here, i met a use case where i would like to have the dropdown in the center below a button (that is itself in the center of the page) would be lovely! 

>> display edit-mode-buttons as column to take less space (feature parity) (would need to remove absolute position of class autocomplete element)

>If you mean having the button go below the input, then it's addressed by a2c50b9

yes, that

>If you mean having edit-mode-buttons themselves displayed on a column, that doesn't seem to be the case on the pre-rewrite version either, see screenshot:
>edit-mode-buttons

jum-s:
- when i enter year field with 1, then i enter a valid date, flash is not removed

- when i enter an invalid entity value (ie. "@foo" in a mastodon field) then click on "save" (flash is displayed): 
 	- then i click on the trash icon, flash message has changed to "missing parameter in body: old-value or new-value"
 	- then i click on the close icon, the invalid value is displayed, as it seems to have successfully updated the entity (but it hasnt, as expected)

- when i create an edition from a work, i got confused about the 1px black border around the title claim. Only later, when i scrolled at the bottom of the page i found the info flash ("Required properties are missing:"). I understand the intention of not focus too much on a flash at first sight of the page, but this border without visible notice was confusing to me. (maybe with a feature parity of "add more details" button, the flash could be more visible? or only display the border when bottom page is first reached?)

- a11y: 
	- the `add-value` button:
		- have a low contrast (#959595 vs white)
		- have a hidden text by default and dont have an aria-label (I dont know how screen readers would render this, but just in case, an `aria-label`)



non-blocking

- when adding a new/non-existing claim, the trash icon could be hidden, as it has the same behavior as the close button on its left.
- on a small screen (412px), when i enter a publisher string, the create button is displaying only "Create a new publish..." (error or feature?)
- when i edit an existing entity value, the suggestions are the same as when adding a new entity value (authors connected to the work somehow), but i expected to have suggestions based on the entity value string in question. I dont know if im clear: when clicking on "Victor Hugo", suggestion is lets say "Gilles Deleuze", but i would like the suggestions as if i typed the string "Victor Hugo" (ie. so suggestions would include "Victor Hugo Valiati"), for easier deduplication i guess.
- when i edit an existing collection title, then i click on save, the title value displayed on the top part of the page is not updated (need to refresh page)
- when i enter a date "1999", then click on the trash icon, then click on cancel, date is reset to "2020" (no need to display a cancel button for date then?) 
- when i enter year field with 1, flash error is unprecise ("invalid property value") could be "year is out of range/too old"
- when i unsuccessfully search for a serie, i would like to click on a button "add a new serie"

maxlath:
> it could focus on the text input (that's fine), but suggestion focus could come back to the first suggestion

done in 2175f58

> actions menu is not visible

fixed in 5a77d4b

> i met a use case where i would like to have the dropdown in the center below a button

done in 817123a

> when i enter year field with 1, then i enter a valid date, flash is not removed

fixed by d9fcb9d

> when i enter an invalid entity value (ie. "@foo" in a mastodon field) then click on "save", then i enter a valid value, flash is not removed.

fixed by 28f2711

> when i enter an non-existing invalid entity value (ie. "@foo" in a mastodon field) then click on "save", flash is not displayed

fixed by f368fbd

> when i create an edition from a work, i got confused about the 1px black border around the title claim. Only later, when i scrolled at the bottom of the page i found the info flash ("Required properties are missing:"). I understand the intention of not focus too much on a flash at first sight of the page, but this border without visible notice was confusing to me. (maybe the flash could be more visible? or only display the border when bottom page is first reached?)

notice added in 2bcd9f6

> a11y : the `add-value` button have a low contrast (#959595 vs white)

fixed by 0c64fc6

> when adding a new/non-existing claim, the trash icon could be hidden, as it has the same behavior as the close button on its left.

fixed by 5f01cf5

> on a small screen (412px), when i enter a publisher string, the create button is displaying only "Create a new publish..." (error or feature?)

fixed by 759f629

> when i edit an existing entity value, the suggestions are the same as when adding a new entity value (authors connected to the work somehow), but i expected to have suggestions based on the entity value string in question. I dont know if im clear: when clicking on "Victor Hugo", suggestion is lets say "Gilles Deleuze", but i would like the suggestions as if i typed the string "Victor Hugo" (ie. so suggestions would include "Victor Hugo Valiati"), for easier deduplication i guess.

I can not reproduce the problem: when editing an existing claim value, `searchText` is set to `currentEntityLabel`, thus it should fetch suggestions based on that label, rather than attempting to get default suggestions (which would look for missing authors from the serie)

> when i edit an existing collection title, then i click on save, the title value displayed on the top part of the page is not updated (need to refresh page)

fixed by 0348d4fb7


> when i enter a date "1999", then click on the trash icon, then click on cancel, date is reset to "2020" (no need to display a cancel button for date then?)

can't reproduce (might have been fixed in the meantime?)

> when i enter year field with 1, flash error is unprecise ("invalid property value") could be "year is out of range/too old"

let aside for now

> when i unsuccessfully search for a serie, i would like to click on a button "add a new serie"

That should already be the case

jum-s (approved):
>> i met a use case where i would like to have the dropdown in the center below a button

>done in 817123a

<3

Issue #253, opened 2021-01-02
create shelves from importers


jum-s:
When someone use the external sites importers, it could create new shelves from the csv given.

i would leave the attempt to resolve on existing shelves, for a future version

We could start by goodread since it was mentioned by [arkhi](https://mamot.fr/@arkhi/105486197441627896)

jum-s:
I am afraid importing shelves from an import file will be quite some work considering the strict schema that candidate items need to fit into
A simpler alternative would be to add the shelves menu already developed for other "add items options"
current view : 
![2021-01-19_17-57](https://user-images.githubusercontent.com/5363918/105067614-c35a1180-5a77-11eb-86e8-59854112995c.png)

other import ways :
![2021-01-19_17-58](https://user-images.githubusercontent.com/5363918/105067707-e389d080-5a77-11eb-88bd-fee0df16a645.png)

this would be trivial and allow a user to customize their csv into multiples files (one per shelf) before importing them

arkhi:
Hello…

I don’t know the code nor the logic behind Inventaire so I don’t understand its inner complexity. My understanding is the following could be done:

1. _Get all arrays of shelves in the export file_ (one array of shelves per book, from what I saw in the csv I have);
1. _Merge those arrays together_ (unique names);
1. _From the resulting merge, unless a shelf with the same name already exists in the user shelves, add all shelves to the user shelves_;
1. _Assign each book to its related shelves_.

I can only mention I doubt users, unless they’re the computer‑nerd **and** book‑nerd types, would spend time curating their exports before importing them. That being written, it won’t hurt to implement the feature it if it’s trivial. :)

jum-s:
We are closer as ever to solve this issue :) 
Your description of the possible implementation @arkhi is quite automatic. Maybe it would be nice for the user to let them choose which "imported shelf" shall be added to which "inventaire shelf".

Proposition: if some "importer shelf" are found, display a checkbox : `Would you like to import your [insert website name]'s shelves too ?`. When checked, a select menu would be display an  "imported shelf" and then a checkbox for each inventaire shelf
something like:

Add [importer shelf name] to the following shelves

- [ ] an inventaire shelf
- [ ] another inventaire shelf
- [ ] create a new shelf with title: **[importer shelf name]**

additional optional feature: default `checkbox: true`  if importer shelf and inventaire shelf have same name

jum-s:
this is now in production :tada: 
ping @arkhi

PR #316, opened 2022-03-08
import shelves

jum-s:
Solving #253
This is a make it work implementation, which is automatically creating new shelves, if some external shelves are found in the importer.

Some choices made:
- No successful results are displayed about the created shelf
- All new shelves are created with sepatrated requests but fired at once (assuming not many are created at once), but could be created sequentially if feels necessary


jum-s:
to be merged once new #322 is done

maxlath:
this can be rebased on master now that #322 has been merged

jum-s:
This is a make it work implementation to import shelves from external websites.

It allow users to automatically create new shelves, based on data import. There is no disambiguation between existing shelves and new ones (one may suggest to dump automatically imported shelves items into existing ones, if shelves titles are the same for example, but this can come later for now). The next step could be to have a tool to easily move/duplicate items from a shelf to another.

When creating a new shelf, visibility is set to private by default, a visibility selector by shelf could be added to the interface, but it might overcrowed the interface, as it is already tricky to present a shelf inport within the global items import (add books). Maybe decoupling the two (shelf/items import) later might make sense, depending of users feedback.

maxlath (approved):

PR #318, opened 2022-03-13
improve entities homonyms search

maxlath:
Currently, a dataadmin going on an entity layout (ex: https://inventaire.io/entity/wd:Q181659) is presented with a list of homonyms made of the exact search results associated with only one label (ex: "Ursula K. Le Guin").

This PR proposes to instead search for more terms when available (ex: "Ursula Kroeber Le Guin", "Ursula Le Guin", "Le Guin", etc), letting the dataadmin the chance to find less straight-forward duplicates, such as:
* https://inventaire.io/entity/inv:783a4084e477b0df0a9e30ec5307875e
* https://inventaire.io/entity/inv:9b970c316e70fd16bcf9933e8564ac6f
* https://inventaire.io/entity/inv:f38d6bb78ae4af173a3a9708d7908393
* https://inventaire.io/entity/inv:b0fe2d7f5775e3678709b7040842f879

This PR removes tasks from the homonyms as the added value there was low: only the presence of an external occurrence was providing an information, but tasks with occurrences appearing high in the task layout, they are often processed before having the time to appear in a random visit on the homonyms section: I don't recall seeing a single task with an external occurrence there

### DIFF
--- a/app/modules/entities/lib/show_homonyms.js
+++ b/app/modules/entities/lib/show_homonyms.js
@@ -1,6 +1,7 @@
 import preq from '#lib/preq'
 import Entities from '../collections/entities.js'
 import loader from '#general/views/templates/loader.hbs'
+import getBestLangValue from './get_best_lang_value'
 
 export default async params => {
   if (!app.user.hasDataadminAccess) return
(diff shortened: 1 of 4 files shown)


jum-s (approved):
to be rebased,
note that the default show_homonyms function is now used by the work entity `entity-layout-rewrite` accepting an entity object instead of a backbone model.

PR #319, opened 2022-05-05
[WIP] Svelte entity layout

jum-s:
Entity layouts replacement with svelte.
Starting by Edition and Work
ItemsList and ItemsMap are shared among the two aforementioned layouts.

specific to leaflet map:  reuse current implementation by 
- loading styles with `getLeaflet` 
- reuse markers styles
but (since involve debackbonify) reuse item and user models, to reuse user's network

there was an attempt to create a svelte modal documented in the branch `svelte-modal`, but without touching <main> it was quite a pain

main difference with feature parity:
- no modal (yet)
- filters by transaction
- initalize map on nearby items not every items
- button 'show everything' to display every items on map

todo:
 - works list in edition layout
 - works claims
 - navigate back to previous entity page when opening item modal 

testing:
- script available on server branch `users-fixture` which creates items of the same edition(s) and network friendship (ie. `$npm run populate-users adamsberg`)


jum-s:
close in favor of cleaner commits in #323

PR #322, opened 2022-06-04
implement shelves visibility

maxlath:
The client side of https://github.com/inventaire/inventaire/pull/624

The trickiest bit was definitely the VisibilitySelector component and its inference system. Some examples:
* if `visibility = [ 'public' ]` then `checked = [ 'public',  'friends',  'groups',  'group:a', 'group:b' ]`
* if `visibility = [ 'groups ]` then `checked = [ 'groups',  'group:a', 'group:b' ]`
* if `group:a` is unchecked, then both `public` and `groups` should be removed from the `visibility` array
The good news is that this component should be straight forward to reuse for other document types (`items`, `lists`)


app/modules/shelves/components/shelf_editor.svelte
```diff
@@ -0,0 +1,183 @@
...
+        model.set({ name, description, visibility, color })
+      }
+      flash = { type: 'success', message: I18n('saved') }
+      await wait(800)
+      app.execute('modal:close')
+    } catch (err) {
+      flash = err
+    }
+  }
+  async function delet () {
```
jum-s, line 46:
```suggestion
  async function deleteShelfConfirmation () {
```

app/modules/general/components/visibility_selector.svelte
```diff
@@ -0,0 +1,146 @@
...
+      class:inferred={visibility.includes('public')}
+      >
+      <input
+        type="checkbox"
+        value="groups"
+        on:click={onGroupsClick}
+
+        bind:group={checked}
+      >
+      {I18n('groups')}
```
jum-s, line 105: An explanation besides `groups` would allow lost users between all those checkboxes. Maybe a little icon `info` , with a tooltip ?
i would include the terms "all present and future groups" or any kind of sentence that would represent what checking this box means: setting a flag for all groups ; and not only allowlisting the current groups (listed right below) which is what the current behavior could suggest (checking every groups checkbox when checking `groups`)  .

app/modules/shelves/components/shelf_editor.svelte
maxlath: addressed in 86c28d6

app/modules/general/components/visibility_selector.svelte
maxlath: addressed in b27b44b

jum-s:
the more i use this new visibility selector, the more a "private" checkbox makes sense to me. It would then be an explicit choice, instead of the currently implicit of not having any checkbox checked

could be v2, but i can dig into it

jum-s:
the more i use this new visibility selector, the more a "private" checkbox makes sense to me. It would then be an explicit choice, instead of the currently implicit of not having any checkbox checked

could be v2, but i can dig into it

maxlath:
The problem with a "private" checkbox is that it breaks the selectors logic. The same purpose could be done without this inconvenient by haveing a "you" checkbox, which can not be unchecked, so that when you uncheck everything else, the only visibility remaining is "you". I would be in favor of shipping the current implementation and see the user feedback we get

maxlath:
Moreover, when every box is unchecked, the displayed selected value is "private", confirming what the user might have guessed
Capture d’écran du 2022-08-29 16-26-03

jum-s:
your last comment is largely depending of the information displayed by the consumer of the visibility selector, hence the user might not see "private" every time a selector is displayed

jum-s (approved):

PR #323, opened 2022-06-15
Entity layout rewrite

jum-s:
Entity layouts replacement with svelte.
Starting by Edition and Work
ItemsList and ItemsMap are shared among the two aforementioned layouts.

specific to leaflet map:  reuse current implementation by 
- loading styles with `getLeaflet` 
- reuse markers styles
but (since involve debackbonify) reuse item and user models, to reuse user's network

there was an attempt to create a svelte modal documented in the branch `svelte-modal`, but without touching `<main>` it was quite a pain

main difference with feature parity:
- no modal (yet)
- filters by transaction
- initalize map on nearby items not every items
- button 'show everything' to display every items on map

todo:
 - works list in edition layout
 - works claims
 - navigate back to previous entity page when opening item modal 

testing:
- script available on server branch `users-fixture` which creates items of the same edition(s) and network friendship (ie. `$npm run populate-users adamsberg`)


jum-s:
> broken entity editor ("entity is undefined")

it feels its more of an understanding of what the editor layout want as entry, i just wanted to make a link to url .../history :s I dont have the courage right now to dig into the entity editor, would you mind doing it?

> no items shown on edition layout (only in certain conditions? when starting from that page?)

no items shown on edition layout was a bug and is fixed, maybe conditions could be v2?

> drop more/less details?

dropped on work and edition layout, but keep feature for element list (currenlty edition list)

> items by categories not visible when empty? Empty gives an information: don't try to look further, there is nothing here

empty items section is be visible enough on work and edition layouts

> series/authors should be regrouped in a distinct "work" section

v2?

> missing fixed width on editions list makes the layout jump when going from 0 edition to some editions

fixed by commit message "editionsList: avoid layout jumping"

> ids such as ISBNs could be at the bottom of the layout

most of the app (including edition uris) rely on isbn, having it not so much down the list make sense to me (altough, other identifiers are actually only available in json). But, besides covers, ISBNs are often the only thing that differentiate an edition from another in a work layout editionsList.

> work claims should be in a different section, to not give the wrong impression that editing the edition could fix those

v2?

> h3.categorie-title: weird font size

if by "wierd" you mean too big, decreased from 1.2 to 1.1 in commit message "clean up"

> reduce main user marker opacity

done with "itemsMap: set opacity on main user marker"

maxlath:
This PR now depends on https://github.com/inventaire/inventaire/pull/640

jum-s:
Before merge:

- [ ] hide filters behind an "advanced features" button on medium and small screens

PR #332, opened 2022-08-06
items: replace listing key by visibility keys array

maxlath:
Same as #322 but for items.
Builds on top of #322, and depends on https://github.com/inventaire/inventaire/pull/632 and https://github.com/inventaire/inventaire/pull/637

To make that change from listing to visibility, I had to rewrite many inventory views in Svelte, making this PR a significant contributor to the inventory Svelte rewrite.

The last commits of this PR also includes the bases of a cross-component state management system, based on per-document Svelte stores that further extends the PR scope: I tried to keep it minimal to limit the scope inflation, more work on this subject could be done in a further PR (ex: to better define data flows and prevent that a store value be modified out of the store [Actions, in the sense of Flux Actions](https://www.youtube.com/watch?v=qtxG-aoXXCM&list=PLoKaNN3BjQX3mxDEVG3oGJx2ByXnue_gR&index=43&t=519s))


jum-s:
> The last commits of this PR also includes the bases of a cross-component state management system, based on per-document Svelte stores that further extends the PR scope

I still dont get why this now, in a 3000 lines PR. I feel against the wall to accept foundations of a new architecture without having time to dig into this store system. I did the minimal to understand the code, but commenting/reviewing this important code that could have come later.

Like polyfills, I would appreciate a new PR for this.

maxlath:
As it's purposefully the last commits, the cross-component state management system could be removed from this PR, but putting that branch in prod without it would likely result in a frustrating user experience, as an update in the item_show modal wouldn't update the inventory-browser item list below. Still, if that's a blocking issue, and as we are indeed a bit in a hurry, I could move it to another PR.

Talking of "foundations of a new architecture" is a bit pushing it: that's very much a first implementation that could easily be reverted and replaced by something more thought through in the future, as we will anyway have to think about this kind of problem. It works like this, so I would be in favor of putting it in prod, but that's in no way a request to commit definitely to that architecture: hurray for code plasticity!

maxlath:
I moved the commits related to state management to #334

jum-s:
i lost my first attempt to comment here, so ill be brief:
Its fine for me to keep it in this branch since we already spent time (including synchronously) on coding/reviewing/commenting this new architecture (as new underlying code that will spread elsewhere later, and new foundation too as is might store more things later)
My point was for the future, to have this kind of code in a new PR next time (like polyfills), despite what it brings to the interface, we could have implemented a temporary alternative like reloading the page to update items.

jum-s (approved):

maxlath:
I mistakenly merged the branch in `shelves-visibility` on which it was based, but remerged in master in ab6fefc55

PR #336, opened 2022-09-03
rewrite the global search bar in Svelte

maxlath:


jum-s (approved):
i surprisingly enjoyed doing the review within a single long file, instead of the usual pattern of many small ones.
Lots of little things could be improved (reduce amount z-indexes, search types declaration) but there is no time for that now. 
And bigger ones (store past searches, remove "all" section in favor of checkboxes), but it wouldnt be a rewrite then.

PR #338, opened 2022-09-03
Display list of items with no shelf

maxlath:
Addressing https://github.com/inventaire/inventaire/issues/430


app/modules/inventory/views/inventory_layout.js
```diff
@@ -106,6 +106,16 @@ export default Marionette.View.extend({
     this.waitForShelvesList.then(() => this.scrollToSection('shelfInfo'))
   },
 
+  showItemsWithoutShelf () {
+    const itemsDataPromise = getItemsData('without-shelf')
+    this.showChildComponent('itemsList', InventoryBrowser, {
+      props: {
+        itemsDataPromise,
+        isMainUser: true,
```
jum-s, line 131: only mainUser can see the items without shelf ?

jum-s:
since its not that far from this PR, do you think its worth to fix the following "bug parity" : when i enter the url of another user (ie. `http://localhost:3005/inventory/Liam1234` from the logged user `adamsberg`) the inv browser shown is mine (the adamsberg one). Note that this error only happens when entering the URL, everything is fine when clicking on an adamsberg inventory link somewhere in the app.
Or i open an issue ?

app/modules/inventory/views/inventory_layout.js
```diff
@@ -110,16 +122,36 @@ export default Marionette.View.extend({
...
+    const itemsDataPromise = getItemsData('without-shelf')
+    this.showChildComponent('shelfInfo', ShelfBox, {
+      props: {
+        withoutShelf: true,
+      }
+    })
+    this.showChildComponent('itemsList', InventoryBrowser, {
+      props: {
+        itemsDataPromise,
+        isMainUser: true,
```
jum-s, line 135:
```suggestion
        // Only mainUser can see their items without shelf
        isMainUser: true,
```

jum-s:
for the record, this implementation is realying on the fact that only the currently logged in user (`mainUser`) can see their items without shelf. Which make sense since the context of this feature is mainly to browse and categorize your own large inventory. It has the advantage also of having a static URL `shelves/without`

jum-s (approved):

PR #339, opened 2022-09-04
Display customizable links to available external resources 

maxlath:
The first commits of this PR add the possibility to edit external ids that could not be edited before, such as OpenLibrary or BNF ids.

The next commits use those ids to display links to the corresponding resources. It's a bit rough in the current implementation, it could be made prettier by using the websites icons instead of text(?)

IMO, it makes some sense to have both those features in this PR as they build on the same shared lib: `app/modules/entities/lib/entity_links.js`.


jum-s (approved):
to merge once #323 is.

PR #345, opened 2022-09-24
update custom properties within the entity editor

jum-s:
solving #346

### DIFF
--- a/app/modules/entities/components/editor/entity_edit.svelte
+++ b/app/modules/entities/components/editor/entity_edit.svelte
@@ -38,7 +38,7 @@
           {favoriteLabel}
         </a>
       </h2>
-      <p class="type">{I18n(entity.type)}</p>
+      <p class="type">{I18n(type)}</p>
       <p class="uri">{uri}</p>
     </div>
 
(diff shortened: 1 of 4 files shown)


maxlath:
There are conflicts with master, could you rebase the branch?

jum-s:
done

maxlath:
Instead of that odd "show more properties" button, what about having an entry in the `entity_edit_menu` like "customize editable properties"?

jum-s:
done in 93127258f

maxlath:
I'm having second thoughts on this:
* I wasn't fond of the button at the bottom, having it in the menu makes it actually harder to reach :/
* I wonder if reusing the list of properties displayed on entities layouts to customize the editor and vice versa will not an annoying behavior: for instance, let's say that I found the ID for a work in the Czech national library (P3184); I don't have it in my custom properties list, so I would have to customize my properties once to display it, set the ID, and customize again as I'm not that interested in that property. What about first showing only the properties in the custom properties list, and then have a `WrapToggler` like in the `entity_create` to display all the available properties in that section, but without changing the custom properties list?

jum-s:
I agree it should be more accessible. 

To consider: the initial proposition included a modal, the intent was to signify that one can easily customize props. By leaving only a view in settings, someone may never know that the feature exists.

Question in your scenario, what if no custom props are checked in a category ? (Now, no category title is displayed, nothing hints that more props are available despite the fact that the user have scrolled all the way down there.)

So what about displaying all categories no matter what: 
- folded category if no custom prop exists
- unfolded if some custom props

If unfolded,  always display the wrapper you mention

What about then having a information box at the bottom of the wrapper, that would invite to customize the properties with a links to the settings?

jum-s:
This PR was a nice feature testing discussion, now depreciated in favor of a clean rewrite in #406.

PR #353, opened 2022-10-13
User profile rewrite

maxlath:
Rewrite the user profile in Svelte, and take that opportunity to improve space use on small screen


jum-s:
maybe something can be done for middle screen view, to not leave that much empty spaces around?

app/modules/users/components/user_profile_buttons.svelte
```diff
@@ -0,0 +1,204 @@
...
+      inGlobalModal={false}
+      on:shelfEditorDone={() => showShelfCreator = false}
+    />
+  </Modal>
+{/if}
+
+<style lang="scss">
+  @import '#general/scss/utils';
+  .profile-buttons{
+    margin: 0.5em;
```
jum-s, line 165: there is an annoying case: the 600px one, which excludes both large and small screen `@media screen` below.
A way to solve this could be to 
```suggestion
    margin: 0.5em;
    margin-left: auto;
```
then to add `margin-left: 0.5em;` to small screen `.profile-buttons`
and to remove the `.profile-buttons` property from large screen

app/modules/users/components/user_profile.svelte
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /*Large screens*/
+  @media screen and (min-width: $smaller-screen) {
+    .avatar-wrapper{
+      flex: 0 0 auto;
+    }
+    .info{
+      flex: 1 0 0;
+      padding: 0.8em 1em;
```
jum-s, line 105: On large but not so large screen (~650) the space below the profile is quite high. I would be for reducing it a bit 
```suggestion
      padding: 0.8em 1em 0 1em
```

maxlath: done in fc3aeba5b

app/modules/users/components/user_profile_buttons.svelte
jum-s: That's correct. Although most of the modified lines in 6ca6c82 are not necessary: the problem was when 2 media queries ranges exclude a screen size. Everything is fine when there is only one media query range.
I would be for defining these `$below-...-screen` variables, but they would only be used for the specific cases of having 2 media queries in the same file.
The DX advantage of only using `$below-...` when necessary (and not systematically), is that when a developer scan the code and find a `$below-...` occurrence, it denotes a specific screen width use case, that one may be careful about updating.

jum-s: Since we disagree on the generalization of this kind or rule and to not block the PR, i propose to add `margin-left: auto;` as suggest above and move 6ca6c82 in another PR ?

maxlath: commit moved to https://github.com/inventaire/inventaire-client/pull/394

PR #354, opened 2022-10-13
Fix user items by entity routing

maxlath:
The `users/:username/inventory/:uri` route was forgotten when moving `/inventory/:username` to `/users/:username/inventory`


jum-s:
If the idea is to transform all `/inventory` routes, it looks like grepping `/inventory/` shows some more occurrences:
- notifications/models/friend_accepted_request.js#L12
- transactions/lib/format_snapshot_data.js#L17
- search/lib/search_results.js#L8

no?

PR #357, opened 2022-10-13
listings: fix /lists/ route

### LINKED ISSUE #352: URL `/lists/` gives an error
while `/lists` is fine.

which seems to happen because the `Listing` router is defined after the `Inventory` one in `/app/init_app.js`. 

Ending up triggering [`'lists/(:id)(/)'`](https://github.com/inventaire/inventaire-client/blob/master/app/modules/listings/listings.js#L8) and not [`'lists(/)'`](https://github.com/inventaire/inventaire-client/blob/master/app/modules/inventory/inventory.js#L30)

maxlath:
fix #352

### DIFF
--- a/app/modules/listings/listings.js
+++ b/app/modules/listings/listings.js
@@ -1,11 +1,12 @@
 import app from '#app/app'
-import { getListingMetadata, getListingPathname, getListingWithElementsById } from './lib/listings.js'
+import { getListingMetadata, getListingPathname, getListingWithElementsById } from '#listings/lib/listings'
 
 export default {
   initialize () {
     const Router = Marionette.AppRouter.extend({
       appRoutes: {
         'lists/(:id)(/)': 'showListing',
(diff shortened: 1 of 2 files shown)


app/modules/listings/listings.js
```diff
@@ -1,31 +1,21 @@
 import app from '#app/app'
-import { getListingMetadata, getListingPathname, getListingWithElementsById } from './lib/listings.js'
 
 export default {
   initialize () {
     const Router = Marionette.AppRouter.extend({
       appRoutes: {
-        'lists/(:id)(/)': 'showListing',
```
jum-s, line 8: dont we lose access to a list by deleting this line ?

jum-s (changes requested):

app/modules/listings/listings.js
maxlath: Indeed, that was a mistake. Fixed in the amended commit bac809abe

Issue #295, opened 2021-03-11
public map on landing page


DougInAMug:
Hello! I heard you didn't have much to do, so I thought I'd write a feature request! :smile: 

I think it would be very attractive to have a public map on the landing page, even if the users are anonymized, to convince non-visionaries it is worth joining.

Right now, you have to create an account just to see if there are any other users in your local area, which is an unfortunately high barrier for people on the edge.

maxlath:
The first version of this public map is online! 🎉 https://inventaire.io/welcome
I would also like to extend a similar mechanism to https://inventaire.io/users/public (see #534). Other ideas are welcome, here or in new issues.

maxlath:
@nclm thanks for reporting, I had the same error, which seemed to be cause by CouchDB being in an unstable state, but I couldn't identify the reason 🤔 Rebooting the server seems to have solved the issue, but please let me know if that happens again

nclm:
It seems to work now indeed :)

PR #359, opened 2022-10-19
welcome layout: rewrite in Svelte and cleanup

maxlath:
Rewriting the landing page layout (aka "welcome" layout) is required to implement #295, and is a good opportunity to do some long overdue cleanup:
- removing mentions: this list is now maintained in the wiki https://wiki.inventaire.io/wiki/Blog (btw, that page could be renamed as that's not a blog, more like a mentions/news feed)
- removing mission: its contextualization role should now be covered in the wiki
- moving credits to the wiki, to make those easier to keep up-to-date
- renaming the link to the wiki as "documentation"

After merging this PR: remove the now obsolete i18n keys


app/modules/welcome/components/landing_screen.svelte
```diff
@@ -0,0 +1,188 @@
...
+      </li>
+      <li>
+        {@html icon('users')}
+        <h3>{I18n('share it with your friends and communities')}</h3>
+        <p>{i18n('Always dreamed of a collaborative library for your coworking / association / book club? Here it is!')}</p>
+      </li>
+      <li>
+        {@html icon('exchange')}
+        <h3>{I18n('Remember the books you lend or borrow')}</h3>
+        <p>{i18n('The books you lend might actually come back now!')}</p>
```
jum-s, line 34: I invoke a change i18n key token (!) to modify the third box `exchange`. This change could be justified by the fact that those texts might be the first and most read words when arriving for the first time on inventaire, the importance of coining some precise words here is quite high. Also, there are more features now than when this `exchange` part was written in 2015, which makes it quite secondary compared to other features from old ones (groups, data editor) to new ones (listings). 

So, I would propose something about the bibliographical data. 

It could be something like `Contribute to an open library` with a Wikdata mention maybe, or more on the book information one can get though inventaire: 
  `Find ideas for the next book you will read`, 
  `Browse users inventories and book lists`. 

If we are open to i18n key changes, then the second box `users` could mention the term `group` like 
  `Make friends, create groups and exchange books together` (which would then wrap up the exchange part).

If this is blocking then i can bury this suggestion in an issue that will be forgotten later.

maxlath: I would indeed prefer to have the rewording in a separate PR, as it is a subject on which we might be nitpicking for a while, blocking the rest of this rewrite, and further developments on this layout. But I do think it's time to start to consider begining a process of getting into rethinking that first screen ([and the time to do it is very soon](https://www.youtube.com/watch?v=l23_-1v__Fg))

jum-s (changes requested):

app/modules/welcome/scss/_welcome_layout_commons.scss
```diff
@@ -0,0 +1,8 @@
+@import '#general/scss/utils';
+
+$welcome-bg-filter: rgba(#123456, 0.5);
+
+h3{
+  margin-top: 0.5em;
```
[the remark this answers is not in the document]
maxlath: done in 32906ed9e, along with other switches from serif to sans-serif for consistency

jum-s (approved):

PR #360, opened 2022-10-20
items_table cosmetics

maxlath:
- remove margin between rows and make rows height more consistent
- reduce the size of the book cover on small screen
- display user information on contexts where that's not an implicitly known fact, typically groups or public views

### DIFF
--- a/app/modules/inventory/components/inventory_browser.svelte
+++ b/app/modules/inventory/components/inventory_browser.svelte
@@ -91,7 +91,12 @@
     <Spinner center={true} />
   </div>
 {:then}
-  <PaginatedItems {Component} {componentProps} {pagination} />
+  <PaginatedItems
+    {Component}
+    {componentProps}
+    {pagination}
+    haveSeveralOwners={groupId != null}
+  />
 {/await}
 
 <style lang="scss">
--- a/app/modules/inventory/components/item_row.svelte
+++ b/app/modules/inventory/components/item_row.svelte
@@ -65,9 +65,8 @@
   .item-row{
     position: relative;
     background-color: #fefefe;
-    @include radius(4px);
-    @include shy-border;
-    margin-bottom: 0.2em;
+    height: 4em;
+    overflow: hidden;
     @include display-flex(row, center, center);
     :global(input[type="checkbox"]){
       padding: 1em;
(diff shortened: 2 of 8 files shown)


app/modules/inventory/components/item_row.svelte
```diff
@@ -33,16 +35,38 @@
...
-      <p class="title">{title}</p>
-      <p class="authors">{authors || ''}</p>
-    </div>
-    <p class="details">{details}</p>
-  </a>
+  <div class="middle">
+    <a
+      href="{pathname}"
+      on:click|stopPropagation={loadInternalLink}
+      class="show-item"
```
jum-s, line 42: when no author, cover is overflown since no minimum height is set on `.show-item`

here is my dev local display, you might find other details than the one mentioned above

jum-s (changes requested):
possibility of clicking o

app/modules/inventory/components/item_row.svelte
```diff
@@ -33,16 +35,38 @@
...
+        {/if}
+      </div>
+      {#if $screen.isLargerThan('$smaller-screen')}
+        <p class="details">{details}</p>
+      {/if}
+    </a>
+
+
+    {#if showUser && $itemStore.user}
+      {@const { username, picture, inventoryPathname } = $itemStore.user}
```
jum-s, line 59: im not sure why this is necessary, why not extracting `user` within reactive block line 22 ?

app/modules/inventory/components/inventory_browser.svelte
```diff
@@ -91,7 +91,12 @@
     <Spinner center={true} />
   </div>
 {:then}
-  <PaginatedItems {Component} {componentProps} {pagination} />
+  <PaginatedItems
+    {Component}
+    {componentProps}
+    {pagination}
+    itemsBelongToMultipleUsers={groupId != null}
```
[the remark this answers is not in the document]
maxlath: I'm not necessarily convinced that it's worth the cost in readability to drop the subject to save a few characters, but I'm not strongly opiniated either, so done in 1521817d

app/modules/inventory/components/item_row.svelte
maxlath: done in 8ac08b2

app/modules/inventory/components/item_row.svelte
maxlath: done in f14750f9e

jum-s (approved):
needs a rebase conflict resolution

Issue #361, opened 2022-10-20
Other editions of multi-work editions


AutomCoding:
Editions of more than one work now only displays the box containing other editions for one of the works.

Examples:
* https://inventaire.io/entity/isbn:9780739451939
* https://inventaire.io/entity/isbn:9780739476581

PR #362, opened 2022-10-21
edition layout: fetch multiple works edition

jum-s:
address #361 

Since multiple works edition is a small fraction of the total amount of editions, and since a work with subentities might be a big chunk of data, the advantages of mutualizing works requests looks less important than the simplicity of having one work subentities request per component.

### DIFF
--- a/app/modules/entities/components/layouts/other_editions.svelte
+++ b/app/modules/entities/components/layouts/other_editions.svelte
@@ -6,13 +6,12 @@
   import EntityImage from '../entity_image.svelte'
   import { getSubEntities } from '../lib/entities'
 
-  export let currentEdition, workUri
+  export let currentEdition, work
 
   let editions = []
   let otherEditions
-
+  const { uri: workUri, label: workLabel } = work
   const getEditionsFromWork = async () => {
-    // TODO: fetch multiple works edition
     editions = await getSubEntities('work', workUri)
     const allOtherEditions = editions.filter(filterEditionWithCover(currentEdition))
     otherEditions = allOtherEditions.splice(0, 4)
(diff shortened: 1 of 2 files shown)


app/modules/entities/components/layouts/other_editions.svelte
```diff
@@ -52,21 +46,22 @@
...
       />
-	  </div>
+    </div>
 	{/if}
 {/await}
 
 <style lang="scss">
   @import '#general/scss/utils';
-  .other-editions{
+  .other-edition{
```
maxlath, line 59: The plural was fine no? It's still other editions from one of the edition works
```suggestion
  .other-editions{
```

app/modules/entities/components/layouts/other_editions.svelte
```diff
@@ -31,12 +30,7 @@
   </div>
 {:then}
 	{#if isNonEmptyArray(otherEditions)}
-	  <div class="other-editions">
-	    <div class="editions-list-title">
-	      <h5>
-	        {I18n('other editions')}
-	      </h5>
-	    </div>
+    <div class="other-edition">
```
maxlath, line 33: Those dives just stack vertically, not using the available horizontal space on large screens. Could those `div.other-editions` be regrouped in a flex-row-wrapped list?
```suggestion
   <ul class="works-other-editions">
      <li class="other-editions">
```
```css
.works-other-editions{
  @include display-flex(row, baseline, center, wrap);
}
```

maxlath (changes requested):

app/modules/entities/components/layouts/other_editions.svelte
jum-s: done in 4f9a137

app/modules/entities/components/layouts/other_editions.svelte
jum-s: done in 083836a

maxlath (approved):

PR #367, opened 2022-11-21
rewrite users_home network/public navs in Svelte

maxlath:
The big chunk here being the rewrite of the public nav map, for which I experimented with ideas from  https://imfeld.dev/writing/svelte_domless_components and https://github.com/ngyewch/svelte-leaflet to be able to use a more Svelte-idiomatic syntax


app/modules/map/lib/config.js
```diff
@@ -8,7 +8,7 @@ export default {
     attribution: `Map data &copy; <a href='http://openstreetmap.org'>OpenStreetMap</a> contributors,
 <a href="http://creativecommons.org/licenses/by-sa/2.0/">CC-BY-SA</a>,
 Imagery © <a href="http://mapbox.com">Mapbox</a>`,
-    minZoom: 2,
+    minZoom: 6,
```
jum-s, line 11: This is problematic if items are displayed on a global scale. Lets say main user position is in indonesia, and there is an item in peru, map bounds will then be exactly in between those two marker locations (main user marker and user item marker), with a 6 zoom, so basically a map that show nothing.
I see two workarounds: 
- either the map does not display markers further than x km from main user (but that too bad since inventaire have this information) 
- or minZoom is 2 as default and a component variable could adjust it if necessary.

jum-s (changes requested):

jum-s:
I try to display a marker on the public map (/users/public), so there is a user poetically named "aaa", who has a publicly shared "Les misérables" item and a position in Paris. After logging with the user "adamsberg" and going to users/public in paris, then "aaa" is not visible, but it should, right? 
Precision: aaa and admasberg are not friends and do not belong to any common group
What am I missing?

app/modules/map/lib/config.js
maxlath: My primary motivation was that when then map is too zoomed-out, only partial results might be displayed anyway, but I didn't think of the case you're referring to. Reverted in fa8287286

maxlath: My primary motivation was that when then map is too zoomed-out, only partial results might be displayed anyway, but I didn't think of the case you're referring to. Reverted in e1565cdbb

app/modules/users/components/network_users_nav.svelte
```diff
@@ -0,0 +1,174 @@
...
+      </div>
+    </div>
+    {#if users}
+      <UsersHomeSectionList docs={users} type="users" />
+    {/if}
+  </div>
+
+  <div class="list-wrapper">
+    <div class="list-label-wrapper">
+      <h2 class="list-label">{I18n('groups')}</h2>
```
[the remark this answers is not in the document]
maxlath: done in 337119936

app/modules/map/lib/map.js
```diff
@@ -110,7 +117,7 @@ export function showUserOnMap (map, user) {
 
   if (user.hasPosition()) {
     const marker = map.addMarker({
-      objectId: user.cid,
+      objectId: `u${user.id}`,
```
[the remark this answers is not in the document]
maxlath: The docs uuids are assumed to be collision-resistant, but I do find it useful to namespace those ids, as we can have several types of objects of a map (typically users and groups): if I see  `u${user.id}` in an error stack trace, without more context, I would be able to directly know that its about a user marker. To ease the pain of your future self, would that work to make the namespace more verbose?: `user-${user.id}`

app/modules/users/components/network_users_nav.svelte
```diff
@@ -0,0 +1,174 @@
...
+      }
+    }
+  }
+
+  /*Small screens*/
+  @media screen and (max-width: $small-screen) {
+    .list-label{
+      margin-bottom: 0.5em;
+      font-size: 1.2em;
+      font-weight: bold;
```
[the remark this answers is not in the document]
maxlath: removed in e3650de6c

app/modules/users/components/users_home_section_list_li.svelte
```diff
@@ -0,0 +1,107 @@
...
+    userIsGroupAdmin = group.admins.find(({ user }) => user === doc._id) != null
+  }
+
+  const onClick = SelectInventoryOnClick({ type, doc })
+</script>
+
+<a
+  class="nav-list-element"
+  href={pathname}
+  on:click={onClick}
```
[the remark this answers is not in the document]
maxlath: yes, that's lazy naming ;) fixed in 2e3dca015

maxlath:
> "aaa" is not visible, but it should, right?

yes, it should: any user with a position should be visible on the map

app/modules/map/lib/map.js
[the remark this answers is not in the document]
jum-s: > would that work to make the namespace more verbose?: `user-${user.id}`

That would be great

jum-s:
> > "aaa" is not visible, but it should, right?
> 
> Yes, it should: any user with a position should be visible on the map. Do you see the user doc in the response from
> 
> ```
> curl 'http://localhost:${DEV_SERVER_PORT}/api/users?action=search-by-position&bbox=[2.1588,48.7608,2.6189,48.9699]' 
> ```
> 
> ? Do you see such a `GET /api/users?action=search-by-position` request being made by the client when you browse the map over paris?

works fine

app/modules/users/components/lib/navs_helpers.js
```diff
@@ -0,0 +1,14 @@
...
+import { isOpenedOutside } from '#lib/utils'
+
+export const isNotMainUser = user => user._id !== app.user.id
+
+export const SelectInventoryOnClick = ({ type, doc }) => e => {
+  if (isOpenedOutside(e)) return
+  let model
+  if (type === 'user') model = new User(doc)
+  else if (type === 'group') model = new Group(doc)
+  app.vent.trigger('inventory:select', type, model)
```
jum-s, line 12: When i click on a user marker, it scrolls to the top tabs.
But when i click on a group marker it scrolls to the inventory

jum-s:
a quite annoying bug to debug and to test happens after navigating the map for a while. The users category is not up to date after moving around on the map
Here is an example in screenshots, it quite easy to reproduce for me if you have quite some marker on the map

Heres the initial state:



Here is the bug : Tomohito should appear in the user category, but instead the main user adamsberg is displayed but it should not. FYI it looks like Tomohito never appears again in the user category no matter how much i zoom out


And actually, if i zoom in and zoom out again, a lot less markers are on the map (aka initial state is not recovered)

app/modules/users/components/network_users_nav.svelte
```diff
@@ -0,0 +1,164 @@
...
+        </button>
+      {/if}
+      <div class="buttons groupMenu">
+        {#if showGroupsMenu}
+          <button on:click={app.Execute('show:groups:search')}
+          >
+            {@html icon('search')}
+            {I18n('search')}
+          </button>
+          <button on:click={app.Execute('show:groups:nearby')}
```
jum-s, line 80: Im not sure of the step to reproduce, but when i m on the "friends & groups" tab and i click on "find on map" of the "

groups" section, and then if i dezoom a lot (around zoom 3 i would say) then the following flash error appears:



note that this does not happen when click on find on map from the friends section

app/modules/users/components/lib/navs_helpers.js
maxlath: might have been fixed by 0a938b0

maxlath:
> After navigating the map, users category is not up to date

should have been fixed by 30945cb2e

maxlath:
> invalid bbox

Should have been fixed by 66044db3e and 54a3da145, depends on server PR https://github.com/inventaire/inventaire/pull/672

app/modules/users/components/network_users_nav.svelte
maxlath: Should have been fixed by 66044db3e and 54a3da145, which depend on server PR https://github.com/inventaire/inventaire/pull/672

jum-s (approved):

Issue #368, opened 2022-11-25
Duplicate shelves when importing multiple times


arjanboltjes:
I recently imported a list of 800+ plus books, from a LibraryThing `.json` file.
The first time the majority of the books were imported, as mentioned on Mastodon [here](https://mastodon.social/@tinyspheresof/109392904432068220). However, some weren't, and after a quick fix and a patch from your side, I imported the same list a few more times.

Inventaire nicely only selected the books that had not been added yet from my file, yet.
However, it also duplicated shelves read from the newly added books, while the same shelves were already in place based on the earlier import.

It would be nice:
1. if this behaviour at import is fixed
2. if we can merge existing shelves

jum-s:
Hi, thank you for reporting.

Since the current state of your items/shelves comes from a patch/fix from our side, may i propose a hacky way to get there without touching the actual code : deleting the old shelf and delete all its items too (youll be asked when deleteing the shelf) and reimporting a fresh new shelf?
It would make it easier for the developers, than entering the blurry area of merging lists of elements which can be tricky.

maxlath:
Alternatively to deleting everything, there is a way to manually merge shelves:
- Go to a shelf that needs to be merge
- Select the "Table" display mode
- Use the menu at the bottom of the screen to "Select all" the items in this shelf, and "Edit selection" > "Edit items shelves"
- Check the shelves in which the current shelf should be merged and "Save"
- (Check that the items were indeed added to the other shelf)
- Use the "edit shelf" menu to delete the shelf

Still a hack, but indeed, we can hope that this kind of problem doesn't occur that often, and thus we can work on more urgent matters for now(?). Let's keep this issue open to see if other people encounter this problem.

arjanboltjes:
Thanks for the replies. I indeed in the meantime did what @maxlath suggested, instead of removing and re-importing.
This worked nicely, although the batch 'edit shelf' is overwriting the old shelves, by the looks of it.
In due time it would be nice to have the availability to add a shelf to a selection without touching the other shelves.

Mostly fixed for me.
I'll leave the issue open, as requested.

maxlath:
Following #369, a basic deduplication is performed: if a shelf already exist with that exact same name, it will be reused, instead of creating a new shelf.

PR #369, opened 2022-11-25
import_items_section: try to find an existing shelf with a matching name before creating a new one

maxlath:
Addressing https://github.com/inventaire/inventaire-client/issues/368

It actually doesn't require that much additional complexity


jum-s (approved):
i guess `external_shelves_importer_description` can mention this new behavior

PR #376, opened 2022-12-22
add window.visualViewport polyfill

maxlath:
To prevent the app from crashing in Firefox < v91


maxlath:
Fast-tracking this to not let the user blocked with a spinner hang

PR #377, opened 2022-12-28
Change svelte linter

maxlath:
Replacing the official but not so developed [`sveltejs/eslint-plugin-svelte3`](https://github.com/sveltejs/eslint-plugin-svelte3) with the much more powerful  [`ota-meshi/eslint-plugin-svelte`](https://github.com/ota-meshi/eslint-plugin-svelte), bringing a good bunch of auto-fix to harmonize the syntax of Svelte components :yum: 

The proposed configuration is done in the first commit, while its effect can be seen in the second commit: I activated many non-recommended [rules](https://ota-meshi.github.io/eslint-plugin-svelte/rules/), but some could maybe be disabled if controversial.


jum-s:
After quite some time trying to configure [prettier](https://prettier.io), i can say that their non-modifiable opinionated configuration is bad (im not the only one [complaining](https://github.com/prettier/prettier-vscode/issues/352) ).

So instead, the last commits setup and fix with stylelint. Rules are soft, as CSS readability is less important than JS).

More exploration report: make it fast option: when benchmarking eslint (prepend `TIMING=1` to the command), the process ends up showing that the `compat` plugin takes a while, maybe it could be disabled in development?

Example on a quite big file:

```
$ TIMING=1 eslint app/modules/entities/components/layouts/works_browser.svelte
Rule                                 | Time (ms) | Relative
:------------------------------------|----------:|--------:
svelte/valid-compile                 |   267.794 |    46.8%
compat/compat                        |   196.162 |    34.3%
indent                               |    31.995 |     5.6%
svelte/indent                        |    16.262 |     2.8%
...
```

jum-s (approved):

jum-s (approved):

PR #379, opened 2023-01-01
Ignore compat

jum-s:
This is an attempt to not extend `compat` plugin, introduced by 170620efd, for a quicker linting (ie. in editor).

This is because, when benchmarking eslint (prepend `TIMING=1` to the command), the process ends up showing that the `compat` plugin takes a while, maybe it could be disabled in development?

Example on a quite big file:

```
$ TIMING=1 eslint app/modules/entities/components/layouts/works_browser.svelte
Rule                                 | Time (ms) | Relative
:------------------------------------|----------:|--------:
svelte/valid-compile                 |   267.794 |    46.8%
compat/compat                        |   196.162 |    34.3%
indent                               |    31.995 |     5.6%
svelte/indent                        |    16.262 |     2.8%
...
```

Unfortunetely, eslint is complainging that a code comment in `viewport.js` wants this plugin. I havent found a workaround for it, hence the WIP status of the PR

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -15,13 +15,6 @@ module.exports = {
   extends: [
     // See https://github.com/standard/eslint-config-standard/blob/master/eslintrc.json
     'standard',
-    // Helps to spot unsupported features that woul result in
-    // babel including the corresponding core-js polyfills
-    // in the bundle served to all users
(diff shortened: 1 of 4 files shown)


maxlath:
The whole `eslint-disable-next-line compat/compat` comment could be removed, as `and_uc 12` is now out of the list of supported browsers

PR #382, opened 2023-01-05
Move loading message within selector's dropdown

jum-s:
Since it can take some time to load facets, and information on what to select can be displayed as soon as advanced option is clicked on.

To do this, we extracted the sync part out of `getWorksFacets`, and allow selector to have a default value placeholder

A friendly message is displayed within the selector (reviewer: don't be lured by the absence of i18n key, CSS should be correct when key exists)

This is preparing ground for adding large reverse claims values to get filtered (ie. wdt:P69 educated at, or wdt:P166 award received)

### DIFF
--- a/app/modules/entities/components/layouts/author.svelte
+++ b/app/modules/entities/components/layouts/author.svelte
@@ -72,31 +72,30 @@
         {/await}
       </div>
     </div>
-    <MissingEntitiesMenu
-      waiting={waitingForSubEntities}
-      questionText="A series or a work by this author is missing in the common database?"
-      {createButtons}
-    />
-    <div class="relatives-lists">
-      <RelativeEntitiesList
-        {entity}
-        claims={entity.claims['wdt:P737']}
-        label={i18n('authors_or_works_influencing_author', { name: entity.label })}
+    <!-- waiting for subentities to not display relative entities list before work browser -->
+    <!-- to not having to push them down while work broser is being displayed -->
+    {#await waitingForSubEntities}
+      <MissingEntitiesMenu
+        waiting={waitingForSubEntities}
+        questionText="A series or a work by this author is missing in the common database?"
+        {createButtons}
       />
-      <RelativeEntitiesList
-        {entity}
-        property={[ 'wdt:P2679', 'wdt:P2680' ]}
-        label={i18n('editions_prefaced_or_postfaced_by_author', { name: entity.label })}
-      />
-      {#each getRelativeEntitiesProperties(type) as property}
+      <div class="relatives-lists">
         <RelativeEntitiesList
           {entity}
-          {property}
-          label={getRelativeEntitiesListLabel({ property, entity })}
+          property={[ 'wdt:P2679', 'wdt:P2680' ]}
+          label={i18n('editions_prefaced_or_postfaced_by_author', { name: entity.label })}
         />
-      {/each}
-    </div>
-    <HomonymDeduplicates {entity} />
+        {#each getRelativeEntitiesProperties(type) as property}
+          <RelativeEntitiesList
+            {entity}
+            {property}
+            label={getRelativeEntitiesListLabel({ property, entity })}
+          />
+        {/each}
+      </div>
+      <HomonymDeduplicates {entity} />
+    {/await}
   </div>
 </BaseLayout>
 
--- a/app/modules/entities/components/layouts/works_browser.svelte
+++ b/app/modules/entities/components/layouts/works_browser.svelte
@@ -32,38 +32,43 @@
 </script>
 <Flash state={flash} />
 <div class="works-browser">
-  <div class="wrapper" class:unwrapped={showControls}>
-    {#if $screen.isSmallerThan(smallScreenThreshold) && isNotEmpty}
-      <button
-        class="toggle-controls"
-        on:click={() => wrapped = !wrapped}
-        aria-controls="works-browser-controls"
-      >
-        {#if showControls}
-          {@html icon('caret-up')}
-        {:else}
-          {@html icon('cog')}
-          {i18n('Advanced options')}
-          {@html icon('caret-down')}
-        {/if}
-      </button>
-    {/if}
-    {#if showControls}
-      {#if isNonEmptyArray(allWorks)}
-        <div class="controls">
-          <WorksBrowserFacets
-            works={allWorks}
-            bind:facets
-            bind:facetsSelectors
-            bind:facetsSelectedValues
-            bind:flash
-          />
-          <WorksBrowserTextFilter bind:textFilterUris />
-          <SelectDropdown bind:value={displayMode} options={displayOptions} buttonLabel={I18n('display_mode')} />
-        </div>
+  {#if isNotEmpty}
+    <div
+      class="wrapper"
+      class:unwrapped={showControls}
+    >
+      {#if $screen.isSmallerThan(smallScreenThreshold) && isNotEmpty}
+        <button
+          class="toggle-controls"
+          on:click={() => wrapped = !wrapped}
+          aria-controls="works-browser-controls"
+        >
+          {#if showControls}
+            {@html icon('caret-up')}
+          {:else}
+            {@html icon('cog')}
+            {i18n('Advanced options')}
+            {@html icon('caret-down')}
+          {/if}
+        </button>
       {/if}
-    {/if}
-  </div>
+      {#if showControls}
+        {#if isNonEmptyArray(allWorks)}
+          <div class="controls">
+            <WorksBrowserFacets
+              works={allWorks}
+              bind:facets
+              bind:facetsSelectors
+              bind:facetsSelectedValues
+              bind:flash
+            />
+            <WorksBrowserTextFilter bind:textFilterUris />
+            <SelectDropdown bind:value={displayMode} options={displayOptions} buttonLabel={I18n('display_mode')} />
+          </div>
+        {/if}
+      {/if}
+    </div>
+  {/if}
 
   {#if sections}
     {#each sections as section}
--- a/app/modules/entities/components/layouts/works_browser.svelte
+++ b/app/modules/entities/components/layouts/works_browser.svelte
@@ -146,13 +146,24 @@
     }
   }
 
-  /* Very small screens */
-  @media screen and (max-width: $very-small-screen){
-    .wrapper:not(.unwrapped){
-      @include display-flex(column, center, stretch);
+  /* Smaller screens */
+  @media screen and (max-width: 450px){
+    .controls{
+      :global(.select-dropdown), :global(.dropdown-content), :global(.works-browser-text-filter){
+        margin: 0.5em;
+        width: 100%;
+      }
     }
     .wrapper{
+      &:not(.unwrapped){
+        @include display-flex(column, center, stretch);
+      }
       @include display-flex(column, center, space-between);
+      &.unwrapped{
+        .toggle-controls{
+          align-self: stretch;
+        }
+      }
     }
   }
 </style>
--- a/app/modules/entities/components/layouts/works_browser.svelte
+++ b/app/modules/entities/components/layouts/works_browser.svelte
@@ -39,11 +39,11 @@
         on:click={() => wrapped = !wrapped}
         aria-controls="works-browser-controls"
       >
-        {@html icon('cog')}
-        {i18n('Advanced options')}
         {#if showControls}
           {@html icon('caret-up')}
         {:else}
+          {@html icon('cog')}
+          {i18n('Advanced options')}
           {@html icon('caret-down')}
         {/if}
       </button>
@@ -84,7 +84,8 @@
     margin-block-end: 0.5em;
     @include radius;
     padding: 0.5em;
-    @include display-flex(row, center, flex-start);
+    @include display-flex(row, space-between);
+    flex: 1;
     :global(.select-dropdown), :global(.dropdown-content){
       inline-size: 10em;
     }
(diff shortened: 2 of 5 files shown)


app/modules/entities/components/lib/works_browser_helpers.js
```diff
@@ -49,6 +52,15 @@ const setWorkFacets = ({ facets, valuesUris, contextProperties }) => work => {
   }
 }
 
+const removeFacetsWithNoKnownValue = facets => {
+  Object.keys(facets).forEach(prop => {
+    const facet = facets[prop]
+    if (notOnlyUnknownKey(facet)) delete facets[prop]
+  })
```
maxlath, line 59: very much non-blocking: this days I prefer this syntax
```suggestion
  for (const [ prop, facet ] of Object.entries(facets)) {
    if (notOnlyUnknownKey(facet)) delete facets[prop]
  }
```

app/modules/entities/components/lib/works_browser_helpers.js
```diff
@@ -3,18 +3,21 @@ import properties from '#entities/lib/properties'
...
   const contextProperties = facetsProperties[context]
   const { facets, facetsSelectedValues } = initialize({ contextProperties })
 
   const valuesUris = []
   works.forEach(setWorkFacets({ facets, valuesUris, contextProperties }))
+  removeFacetsWithNoKnownValue(facets)
+  return { valuesUris, facets, facetsSelectedValues }
+}
 
+export async function getFacetsSelectors (facetsObj) {
```
maxlath, line 16: I think the function name could be more indicative of how it works with the object returned by `getWorksFacets`, and how it's different from it
```suggestion
export async function getWorksFacetsBasicInfoAndSelectors (worksFacets) {
```
maybe?

app/modules/entities/components/layouts/works_browser_facets.svelte
```diff
@@ -1,31 +1,39 @@
...
-        bind:value={facetsSelectedValues[property]}
-        options={facetsSelectors[property].options}
-        resetValue="all"
-        buttonLabel={i18n(property)}
-        withImage={entityProperties.includes(property)}
-      />
-    {/if}
-  {/each}
-{/await}
+  ({ facets, facetsSelectedValues } = facetsObj)
```
maxlath, line 18: non-blocking, but I would like to stick to this convention to have a `;` before object destructive re-assignments that aren't in a one-liner block, as it prevents issues when moving the line around, without having the cognitive load to need to check what is the line above to know if this could mistakenly be interpreted as a function call. 
```suggestion
  ;({ facets, facetsSelectedValues } = facetsObj)
```

app/modules/entities/components/layouts/works_browser_facets.svelte
```diff
@@ -1,31 +1,39 @@
...
+  Promise.all([
+    getFacetsSelectors(facetsObj),
+    displayLoadingInfoMessage(facetsObj)
+  ])
+  .then(res => ({ facets, facetsSelectedValues, facetsSelectors } = res[0]))
+  .catch(err => flash = err)
+  .finally(() => loadingMessage = false)
```
maxlath, line 25:
```suggestion
  waitingForOptions = getFacetsSelectors(facetsObj)
    .then(res => ({ facets, facetsSelectedValues, facetsSelectors } = res))
    .catch(err => flash = err)
```

app/modules/entities/components/layouts/works_browser_facets.svelte
```diff
@@ -1,31 +1,39 @@
...
+</script>
+{#each Object.keys(facets) as property}
+  <SelectDropdown
+    bind:value={facetsSelectedValues[property]}
+    options={facetsSelectors?.[property].options}
+    resetValue='all'
+    {loadingMessage}
```
maxlath, line 32:
```suggestion
    {loadingMessage}
    {waitingForOptions}
```

app/modules/entities/components/layouts/works_browser_facets.svelte
```diff
@@ -1,31 +1,39 @@
...
 
-  const waitForFacets = getWorksFacets({ works, context: layoutContext })
-    .then(res => ({ facets, facetsSelectedValues, facetsSelectors } = res))
-    .catch(err => flash = err)
-</script>
+  let loadingMessage
+  const displayLoadingInfoMessage = async facetsObj => {
+    const urisSize = facetsObj.valuesUris.length
+    loadingMessage = I18n('loading_may_take_a_while', { smart_count: urisSize })
+  }
```
maxlath, line 16: Nothing async here, and see below for why `loadingMessage` could be a `const`. Also, if the message includes `urisCount`, this will probably not be a very generalist message, so having a more specific key (such as `loading_facets_may_take_a_while`) could be helpful:
```suggestion
  const urisCount = facetsObj.valuesUris.length
  const loadingMessage = I18n('loading_facets_may_take_a_while', { smart_count: urisCount })
```

app/modules/general/components/select_dropdown.svelte
```diff
@@ -20,7 +27,8 @@
     e.preventDefault()
   }
 
-  $: currentOption = options.find(option => option.value === value)
+  const optionPlaceholder = { value: 'all', text: I18n('loading') }
```
maxlath, line 30: Having all the dropdown menus displaying "loading" (or the even more verbose "chargement en cours") looks a bit unnecessarily bloated; what about just a spinner? I haven't tried it, but it might be quite simple to implement, by having an `{#if}` block within `.inner-select-option`: `{#if option}...{:else}<Spinner />{/if}`

maxlath (changes requested):
Nice! I like how it reduces elements suddenly jumping at your face after loading

app/modules/general/components/select_dropdown.svelte
jum-s: I dont think clicking on several/all facets selectors while loading is a reasonable usecase, why would one click on everything after they know one facet is loading? (besides trolling/testing, which can/must be ignored). And even if they do, then "bloating" a bit (and temporarily) the interface looks relevant to me, interface is shouting  "wait, please!", more and more, as more facets selectors are displayed.

Besides this example, i will also answer on the argument itself:
When facets are loading, selector input is already displaying a text stating "Loading", as a first attempt to inform that one must wait. I can see 2 reasons why the user would still choose to click anyway: 
- not reading the "Loading" text (ie. knowing the interface, clicking by habit),
- still wanting to explore whats happening, as an attempt to understand why this loading state is so long (and it can really take a solid minute for 1000 entities cf. subjects layout in #383).
I think your suggestion (having spinner without help text) would satisfy the first case above, but if one would like to know why, then inventaire can display a hint (aka. how heavy the entities amount is `urisSize`).

If you are not convinced by this argumentation, it's fine for me to delete this text, as it simplifies the code, but i prioritize "displaying information when available" over "code simplification".

jum-s:
After loading hundreds of times these entities layouts recently, i came to the conclusion that loading `relativeEntitiesList` often take less time. And i think it would be a better UX to only display those `relativeEntitiesList` only after `worksBrowser` is displayed to not have this jump/pushed down. Would you be okay if i add a commit to this PR that does that?

app/modules/entities/components/lib/works_browser_helpers.js
jum-s: knowing #383 introduce more than works type in facets (ie humans and editions) i would like to keep it entity type agnostic, but informing that this function is fetching server side entities totally make sense. I propose `getFacetsEntitiesBasicInfo` in commit 14cb936f3
 
your second suggestion `worksFacets` doesnt reflect enough for me what it is: a custom `Object` having very specific keys, but not the facets themselves.  `worksFacets` sounds like this variable is facets, but its not.

app/modules/general/components/select_dropdown.svelte
```diff
@@ -55,22 +63,35 @@
...
-            role="option"
-            title={option.text}
-            aria-selected={option.value === value}
-            on:click={() => value = option.value}
-          >
-            <SelectDropdownOption {option} {withImage} />
-          </button>
-        {/if}
-      {/each}
+      {#if loadingMessage}
```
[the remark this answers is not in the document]
jum-s: done in 98884209d

app/modules/entities/components/layouts/works_browser.svelte
```diff
@@ -94,7 +99,7 @@
   }
 
   .wrapper{
-    @include display-flex(column, stretch);
+    @include display-flex(row-reverse, space-between);
```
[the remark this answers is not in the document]
jum-s: done in b48b8efa3

app/modules/entities/components/layouts/works_browser.svelte
```diff
@@ -103,9 +108,12 @@
     &.unwrapped{
       background-color: $off-white;
       .toggle-controls{
-        align-self: flex-end;
+        align-self: flex-start;
       }
     }
+    &.empty{
+      display: none;
+    }
```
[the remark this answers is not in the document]
jum-s: no
done in 81dbda77e

app/modules/entities/components/lib/works_browser_helpers.js
jum-s: yeah more obscure for JS newcomers in my opinion (`.forEach` and `.keys` are quite basic compared to array destructuring and `.entries`)
but done anyway in 81dbda77e

app/modules/entities/components/layouts/works_browser_facets.svelte
jum-s: done in 98884209d

app/modules/entities/components/layouts/works_browser_facets.svelte
jum-s: done in 98884209d

app/modules/entities/components/layouts/works_browser_facets.svelte
jum-s: done in 98884209d

app/modules/entities/components/layouts/works_browser_facets.svelte
jum-s: done in 98884209d

app/modules/entities/components/lib/works_browser_helpers.js
maxlath: Come on, it's not some dirty hack, it's a very well documented standard function, now very well established in the JS ecosystem, and giving the possibility to map/filter on non-array objects: that should be on every JS dev toolbox! If you consider it to be obscure, I'm happy to bring some light to it ;)

maxlath:
I rebased this branch on master, resolved the conflicts, and force pushed, but I can revert it if that's not welcome

app/modules/general/components/select_dropdown.svelte
maxlath: Are we talking of the same thing? This is what I have in mind and find a bit too verbose:
Capture d’écran du 2023-02-28 20-38-25

maxlath: > it can really take a solid minute for 1000 entities

Yes, those layouts aren't doing pagination anymore, which might be an issue for publishers in particular. I was considering the possibility to move facets to the server, just like `inventory-view`, but with better caching, which would mean having to properly design the cache invalidation... That's why I would be in favor of not making those layouts pixel perfect just yet, there might still be some heavy work required.

app/modules/entities/components/layouts/author.svelte
```diff
@@ -72,39 +72,41 @@
...
-      waiting={waitingForSubEntities}
-      questionText="A series or a work by this author is missing in the common database?"
-      {createButtons}
-    />
-    <div class="relatives-lists">
-      <RelativeEntitiesList
-        {entity}
-        claims={entity.claims['wdt:P737']}
-        label={i18n('authors_or_works_influencing_author', { name: entity.label })}
+    {#if sections}
```
maxlath, line 75: [non-blocking]
Using an await seems to better communicate why there is this block
```suggestion
    {#await waitingForSubEntities then}
```
That could still deserve a comment though, as there is no clear dependency between `sections` or `waitingForSubEntities` and what's in the block.

app/modules/entities/components/layouts/works_browser_facets.svelte
```diff
@@ -1,31 +1,34 @@
...
+  import { i18n, I18n } from '#user/lib/i18n'
 
   export let works, facets, facetsSelectors, facetsSelectedValues, flash
 
   const layoutContext = getContext('layout-context')
+  const facetsObj = getWorksFacets({ works, context: layoutContext })
 
-  const waitForFacets = getWorksFacets({ works, context: layoutContext })
+  let loadingMessage
+  const urisSize = facetsObj.valuesUris.length
```
maxlath, line 13: In this code base, "size" is used only for fonts and images so far (outside of Elasicsearch DSL), while arrays have "length" and "count". Unless there is a particular reason why "size" is a better fit?
```suggestion
  const urisCount = facetsObj.valuesUris.length
```

jum-s: That's a developer's habit coming from past inventaire life. I used both `size` and `count` interchangeably, but since it hurts your eyes, i changed it in e5a6b238a

app/modules/entities/components/layouts/author.svelte
jum-s: besides this line has been introduced by this PR (but by f505be9ae6) i like how `if sections` documents the fact that `WorksBrowserSection` must have defined `sections` to be displayed, not only subEntities but actually sections. The `waitingForSubEntities` feels more detach from what it actually does, since future maintainer would actually need to understand first that sections is filled with subEntities. `If sections` feels more simple, saying: no sections, no cookies, nothing to display.

Actually im not even sure this `if` or `await` is even necessary, removing it does seem to hurt the display. I removed it in 11b0c9561

jum-s: done in 91d01f25c

app/modules/general/components/select_dropdown.svelte
jum-s: > Are we talking of the same thing?

yes, I misunderstood you here, this it not the behavior i had in mind. Could be because of the rebase? Anyway. What i had in mind was to fakingly display "All", and then if a user is quick enough to click and display the dropdown, then display the loading message with the spinner.

This is done with bce58dec1 but its not what you suggested, would that be okay for you?

maxlath (approved):

PR #389, opened 2023-01-27
profile settings: attempt to discourage spam accounts

maxlath:
addressing https://github.com/inventaire/inventaire/issues/503 by not allowing to save a user bio that contains a URL and a suspect keyword


app/lib/spam.js
```diff
@@ -0,0 +1,27 @@
+export function looksLikeSpam (str) {
+  return urlPattern.test(str) && suspectKeywordPattern.test(str)
+}
+
+const urlPattern = /(http|www\.|\w+\.\w+\/)/i
+
+const suspectKeyword = [
```
jum-s, line 7:
```suggestion
const suspectKeywords = [
```

jum-s:
Quite smart to use status code for this purpose.
What about extracting hardcoded `suspectKeywords` into a `.gitignored` file `.suspect-keywords`, or anywhere none git covered in order to easily update the list without having to commit it ?

maxlath:
@jum-s It makes the setup more complex, but there is indeed the benefit to let the instance administrator adapt the list to their spammer pool. I moved the list to the server config, where it can then be modified as any config parameter https://github.com/inventaire/inventaire/pull/668 What do you think?

app/lib/error.js
```diff
@@ -49,11 +49,13 @@ const error_ = {
...
   // Log and report formatted errors to the server, without throwing
-  report (message, context = {}) {
+  report (message, context = {}, statusCode = 599) {
     context = deepClone(context)
     context.location = location.href
-    // Non-standard convention: 599 = client implementation error
-    const err = formatError(message, 599, context)
+    // Non-standard convention
+    // 599 = client implementation error
+    // 598 = user abuse
```
jum-s, line 57:
```suggestion
    // 598 = suspected user abuse or spam account
```
Non-standard requires being having a precise documentation imo

app/lib/spam.js
```diff
@@ -0,0 +1,27 @@
...
+    } else {
+      suspectKeywordsPattern = null
+    }
+  }
+}
+
+export async function looksLikeSpam (str) {
+  await initSuspectKeywordsPattern()
+  if (suspectKeywordsPattern) {
+    return urlPattern.test(str) && suspectKeywordsPattern.test(str)
```
jum-s, line 21:
```suggestion
    return urlPattern.test(str) || suspectKeywordsPattern.test(str)
```

maxlath: This would mean banning any URL from user profiles, which is way to restrictive imo. There are some very legit links in profile bios.

app/lib/spam.js
```diff
@@ -0,0 +1,27 @@
...
+  } else {
+    return false
+  }
+}
+
+const urlPattern = /(http|www\.|\w+\.\w+\/)/i
```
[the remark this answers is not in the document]
maxlath: An if/else from my initial implementation could indeed be dropped, with the help of the `?.` operator
```js
export async function looksLikeSpam (str) {
  await initSuspectKeywordsPattern()
  return suspectKeywordsPattern?.test(str) && urlPattern.test(str)
}
```
but I would prefer an implementation that doesn't unnecessarily redefine `suspectKeywordsPattern` at every call

app/lib/spam.js
jum-s: ah that's a feature, my bad.

app/lib/spam.js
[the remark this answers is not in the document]
jum-s: indeed, moving `let suspectKeywordsPattern` below `let config` does the trick, no? Then `initSuspectKeywordsPattern` is called only once since `if (!config)` prevents requesting config at every call no?

app/lib/error.js
maxlath: done in 0888cd876

app/lib/spam.js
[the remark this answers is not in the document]
maxlath: I find both our implementation hard to read. What do you think of an implementation relying on an `initialized` boolean to explicit what's happening there? f11895842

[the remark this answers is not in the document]
jum-s: yep

jum-s (approved):

PR #394, opened 2023-02-16
scss: prevent max-width media queries to include the thresold value

maxlath:
to avoid having both rules below and above the threshold to apply at the same time.

max-width and min-width are inclusive comparators (<= and >=), there is unforunately no strict comparators (< and >). An other work around would be with the `not` operator, but that makes queries more verbose and less readeable See https://stackoverflow.com/questions/48601618/strict-inequality-in-css-media-queries

PR extracted from #353, see previous discussion https://github.com/inventaire/inventaire-client/pull/353#pullrequestreview-1196531252

### DIFF
--- a/app/modules/entities/components/editor/claim_editor.svelte
+++ b/app/modules/entities/components/editor/claim_editor.svelte
@@ -186,7 +186,7 @@
       margin-bottom: 1em;
     }
     /* Small screens */
-    @media screen and (max-width: $smaller-screen){
+    @media screen and (max-width: $below-smaller-screen){
       flex-direction: column;
       padding: 0.5em 0.2em;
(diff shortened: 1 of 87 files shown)


maxlath:
There ARE actually strict comparators since [Media Queries Level 4](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries#syntax_improvements_in_level_4), and as the updated [stylelint highlighted with lint errors](https://stylelint.io/user-guide/rules/media-feature-range-notation/). 

In 9ccaa16a0, I replaced the prefix notation by context notation, and introduced the convention to only use 2 operators: `>=` and `<` so that no threshold value can be shared between 2 media queries.

PR #397, opened 2023-03-14
Hide spinner when 100 results are displayed

jum-s:
Since its the maximum displayable. 
This will not give the false impression that more results will be displayed if the user waits.

This is a quick fix based on the fact that the server is not displaying more than [100 results](https://github.com/inventaire/inventaire/blob/master/server/controllers/search/search.js#L18).

### DIFF
--- a/app/modules/search/components/global_search_bar.svelte
+++ b/app/modules/search/components/global_search_bar.svelte
@@ -40,6 +40,9 @@
   let showSearchControls
   let hasMore = false
 
+  // In sync with limit parameter in `server/controllers/search/search.js`
+  const resultsMaxLength = 100
+
   async function search () {
     const searchKey = `${selectedCategory}:${selectedSection}:${searchText}`
     hasMore = false
@@ -88,7 +91,8 @@
         limit: searchBatchLength,
         offset: searchOffset,
       })
-      hasMore = res.continue != null
+      hasMore = res.continue != null && res.continue < resultsMaxLength
+
       return res.results.map(serializeSubject)
     } else {
       // Increasing search limit instead of offset, as search pages aren't stable:
@@ -102,7 +106,8 @@
         search: searchText,
         limit: searchLimit,
       })
-      hasMore = res.continue != null
+      hasMore = res.continue != null && res.continue < resultsMaxLength
+
       return res.results
     }
   }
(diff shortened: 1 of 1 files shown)


app/modules/search/components/global_search_bar.svelte
```diff
@@ -88,7 +91,9 @@
         limit: searchBatchLength,
         offset: searchOffset,
       })
-      hasMore = res.continue != null
+      hasMore = res.continue < resultsMaxLength
```
maxlath, line 94: Fun JS fact I prefer not to have to think about
```js
null < 100 // true
undefined < 100 // false
```
```suggestion
      hasMore = res.continue != null && res.continue < resultsMaxLength
```

app/modules/search/components/global_search_bar.svelte
```diff
@@ -102,7 +107,9 @@
         search: searchText,
         limit: searchLimit,
       })
-      hasMore = res.continue != null
+      hasMore = res.continue < resultsMaxLength
+
+
```
maxlath, line 112: Some linter errors
```suggestion
```

maxlath (changes requested):

app/modules/search/components/global_search_bar.svelte
jum-s: thx, this cannot stick in my head. I made a note to try to not forget. Fixed in 1c22ef762

app/modules/search/components/global_search_bar.svelte
jum-s: Fixed in 1c22ef7

maxlath:
```js
  limit: { default: 10, max: 100 },
  offset: { default: 0, max: 500 },
```
so a single request limit is 100, but more batches can be requested: the effective limit is `maxOffset+maxLimit = 500 + 100 = 600`, no?

jum-s:
>   offset: { default: 0, max: 500 },

Offsets are for `subjects` search only, which does not appear to display more results than the first batch... 

>  but more batches can be requested

Yes, this PR takes for a fact that those requests (beyond 100) do not provide any results (hence keep spinner spinning). As mentioned, this is a quick fix to not lure users into waiting for results that will never come.

In a better world, we would need some investigation to find out why additional batches are not displayed. 

This PR was just a quick fix 6 months ago for a better UI. It could be discarded in favor of the aforementioned investigation. But if not, i would be for a merge.

PR #398, opened 2023-03-24
finish users_home_layout rewrite to Svelte

maxlath:
I had to revert the merge from #367 as after further tests, it was not ready for prod. Namely, the legacy backbone implementation of `users_home_layout` was not doing the transitions properly, stacking the different layouts rather than replacing them, in certain cases.

This PR attempts to finish the job by rewriting the last pieces of that layout to Svelte, most importantly `users_home_layout` itself.

To avoid to the reviewer the need to re-review the parts that were already validated in #367, this PR is based a special branch [`users-layout-rewrite--validated-by-367`](https://github.com/inventaire/inventaire-client/tree/users-layout-rewrite--validated-by-367). Thus, to merge this PR once validated:
- rebase on then `master` and push `users-layout-rewrite`
- change this PR base to `master`
- merge


app/modules/users/components/users_home_section_list.svelte
```diff
@@ -1,17 +1,21 @@
...
 
 {#if hideList}
   <p class="hidden-list-message">{hideListMessage}</p>
 {:else}
-  <ul class:members={type === 'members'}>
+  <ul class:members={type === 'members'} class="users-home-section-list">
     {#each docs as doc (doc._id)}
       <li class="user">
-        <UsersHomeSectionListLi {doc} />
+        <UsersHomeSectionListLi {doc} on:select={bubbleUpComponentEvent} />
```
jum-s, line 18: i guess i let through during the `users-layout-rewrite` review, but the `CurrentEvents` created by `createEventDispatcher`[dont bubble](https://svelte.dev/docs#run-time-svelte-createeventdispatcher) per se (aka to the parent DOM component), here BublingUp refers to the parent Svelte component not the DOM, right? Hence `dispatch` sounds more appropriate?
```suggestion
        <UsersHomeSectionListLi {doc} on:select={dispatchEvent} />
```

app/modules/users/helpers.js
```diff
@@ -30,6 +30,11 @@ export default function (app) {
       return userModel.toJSON()
     },
 
+    async getUsersData (ids) {
```
jum-s, line 33:
```suggestion
    async getUsersModelsFromIds (ids) {
```

app/modules/users/components/inventory_or_listing_nav.svelte
```diff
@@ -1,47 +1,50 @@
 <script>
   import { i18n } from '#user/lib/i18n'
-  import app from '#app/app'
   import { isOpenedOutside } from '#lib/utils'
-  import { isNumber } from 'underscore'
+  // import { isNumber } from 'underscore'
 
-  export let userModel, groupModel, currentSection
+  export let user, group, currentSection
```
jum-s, line 6: Manipulating Backbone model was on purpose, in order to know what we are dealing with here. The idea was to only call models when necessary (the lastet possible), to not depend on models too much.

app/modules/shelves/components/user_inventory.svelte
```diff
@@ -0,0 +1,74 @@
...
+
+{#if selectedShelf === 'without-shelf'}
+  <ShelfBox withoutShelf={true} on:closeShelf={onCloseShelf} />
+  <InventoryBrowser
+    itemsDataPromise={getInventoryView('without-shelf')}
+    {isMainUser}
+  />
+{:else if selectedShelf}
+  <ShelfBox bind:shelf={selectedShelf} on:closeShelf={onCloseShelf} />
+  <!-- Recreate component when selectedShelf changes, see https://svelte.dev/docs#template-syntax-key -->
```
jum-s, line 59: wouh very nice, i needed that for the task rewrite.

app/modules/inventory/components/inventory_browser.svelte
```diff
@@ -94,26 +101,30 @@
...
+    {intersectionWorkUris}
   />
-{/await}
+
+  {#await waitForInventoryData}
+    <div class="spinner-wrap">
+      <Spinner center={true} />
+    </div>
+  {:then}
+    <PaginatedItems
```
jum-s, line 120: If i go to friends & group tab and dont wait long enough before going back to my inventory, then 'Nothing to show here' is displayed instead of my items. Then if i scroll down, items are actually displayed. Only happen on large screen.

app/modules/users/components/inventory_or_listing_nav.svelte
```diff
@@ -1,47 +1,50 @@
...
+  // }
 
   const showSection = (e, section) => {
     if (isOpenedOutside(e)) return
     currentSection = section
-    app.vent.trigger('show:inventory:or:listing:section', { section, userModel, groupModel })
+    if (currentSection === 'inventory') {
+      app.navigate(inventoryPathname, { pageSectionElement: navEl })
+    } else {
+      app.navigate(listingsPathname, { pageSectionElement: navEl })
```
jum-s, line 40: From a listings browser layout, when i click on a listing (aka. i get extracted from user layout context) then, if i click on the browser precedent button, i expect to be back to the listings page, but instead the inventory browser (items) is displayed with a URL `/users/adamsberg/lists`

app/modules/inventory/components/inventory_browser.svelte
jum-s: Would it be a good time to make inventory layout component nested inside its parents (user, groups at least, since public is a bit specific), instead of modifying the state of the inventory broswer every time a tab is changed?

app/modules/users/components/users_home_nav.svelte
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /* Small screens */
+  @media screen and (max-width: $small-screen){
+    #tabs{
+      margin-top: 0.5em;
+    }
+    a{
+      height: 3em;
+      flex-direction: column;
```
jum-s, line 90: breakpoint for this should be lower

app/modules/shelves/components/shelf_box.svelte
```diff
@@ -6,23 +6,38 @@
...
+    app.navigate(pathname, { pageSectionElement: shelfBoxEl })
+  }
+
+  const debouncedOnFocus = debounce(onFocus, 500, true)
+
+  $: if ($focusStore.type === 'shelf') debouncedOnFocus()
 </script>
 
-<div class="shelf-box">
+<div class="shelf-box" bind:this={shelfBoxEl}>
```
jum-s, line 40: somehow `shelf-box` lost its background color, making is difficult to understand its limit when filters are wrapped

app/modules/users/components/users_home_layout.svelte
```diff
@@ -0,0 +1,88 @@
...
+  import { resizeObserver } from '#lib/components/actions/resize_observer'
+
+  export let user, group, shelf, section, subsection = 'inventory'
+
+  const { loggedIn } = app.user
+
+  if (user?._id === app.user.id) section = 'user'
+
+  const focusStore = writable({})
+  setContext('focus-store', focusStore)
```
jum-s, line 22: I thought the consensus obtained after hours of undocumented conversation was to agree on context based variable but to forbid writable stores... and boum few months later, conversation is forgotten and look at what is happening.
I know it makes it easier to write, but side-effects of such `$focusStore = { type: 'shelf', doc: selectedShelf }` (passing the whole document (!)) can get out of control. This ticking bomb is hard to swallow for me.

jum-s (approved):

app/modules/users/helpers.js
maxlath: The "Json" in that name would be propagating the Backbone lie (that `toJSON` method does not return json but an object). But I agree the "Data" part in the name is smelly. The dirty part in this, which is actually the point of reusing that legacy Backbone model based `getUsersModels` function, is that `getUsersModels` builds a local in-memory cache of users. That could indeed be worth making explicit.

app/modules/users/components/inventory_or_listing_nav.svelte
maxlath: You may ask :stuck_out_tongue_winking_eye:

maxlath: We are getting rid of Backbone, so if there is less backbone model, shouldn't it be better? What Backbone-model-free solution would you prefer? There is indeed some implicit here, which is that we are dealing with a serialized user object or group object (as they are expected to have `inventoryPathname` and `listingsPathname` attributes), but that seems acceptable to me. We could get much more explicit by typing our component interfaces with Typescript, but we are not there yet.

app/modules/inventory/components/inventory_browser.svelte
maxlath: The good thing with this rewrite is that those main-user/network sections are largely independent (compared to the previous implementation where the same div was reused to display items for the 2 different sections), so we can confidently rule out that it has to do with the transition from one section to the other.

I was able to reproduce the issue with the main user inventory browser alone (which made debugging much less painful ^^), and apparently fix the problem 2200c9a, but I still would have trouble explaining the exact sequence of events that led to that state ><

app/modules/users/components/inventory_or_listing_nav.svelte
maxlath: fixed by 4e4562a92

app/modules/users/components/users_home_nav.svelte
maxlath: fixed in 26f6a984c

app/modules/shelves/components/shelf_box.svelte
maxlath: fixed in 9ed203d4f

app/modules/users/components/users_home_layout.svelte
maxlath: It's a great dev tool: you share state with all the children, and then when time comes to refactor, you can search which component sets the store value (`$focusStore =`) and which react to it (typically in `$:.*$focusStore`), and if the store and/or context does not seem necessary after all, refactor accordingly. As I now came to refactoring, it appears that the `$focusStore.doc` wasn't even used, so there is no more whole document passing. I tried refactoring without `getContext`/`setContext` in 96f6aac7f: we go from 18 occurrences of the word `focusStore` to 37, making the whole thing more explicit (good) but also more noisy/less grepable (not good). I'm fine going with the explicit version for now, but I don't think that it's such a foot gun that should be locked away: with some conventions on namespacing to guaranty grepability, that solves some real pain points in deeply nested layouts.

app/modules/users/components/users_home_section_list.svelte
jum-s: But what is the diff between bubble up and dispatch? If no difference, why not sticking to svelte naming? (+ fewer keystrokes).

app/modules/users/components/users_home_layout.svelte
jum-s: Im sure you know my position: maybe we dont need deeply nested layouts in the first place, and when we do, better show it everywhere with lots of warnings and line of code than implying that this component is simple, but if you find a variable undeclared in the file the you just need to grep the whole codebase find where things are set. 
It's not a great dev tool for me, it hides complexity, and we should not do that for variables used in a lots of different components, cuz that that sound like a future lengthy debugging (not difficult, but lengthy).

PS: we had that conversation and we agreed that writable stores shall be banned from our codebase, dont remember it? It was at the time we wrote this framasoft article.

app/modules/users/helpers.js
jum-s: or `getSerializedUsers` ? So data naming can apply to other frameworks than bb. 

Its only to signify that this user object is not the user doc coming straight from the server..

app/modules/users/components/inventory_or_listing_nav.svelte
jum-s: It was on purpose to explicitly stating it's a bockbone model. But yes, its better without it.

jum-s:
when i load page `/users/network`, then i click on my username (aka got to my inventory `users/adamsberg/inventory`) my inventory is blank (
Peek 2023-04-23 12-03
with no error message in console).

app/modules/users/components/users_home_layout.svelte
jum-s: Note that im far from being against stores (`$user` is great, although should probably be renamed `$currentUser`) but writable is a different matter. Another fear of mine is that the document passed through the `focusStore`, as it is difficult to find what make `doc` valid (aka how its formatted, what it should/must contain). I would feel like postponing writtable until some typescript is around to type this wild multi-purpose`doc`.

jum-s: Also, no matter of the outcome of this conversation, I would be for renaming `focusStore` -> `focusSection`

app/modules/users/components/users_home_section_list.svelte
maxlath: My understanding of the nuance is that `dispatch` triggers an event on an object ([typically a DOM element in web standard](https://dom.spec.whatwg.org/#events)), while the `bubble` flag determines if that event is automatically forwarded to parent elements of the DOM. Dispatching and forwarding are different in the fact that dispatching sets the object as the event origin (typically the `event.currentTarget` in DOM events), while forwarding just passes the event up without changing the event origin(?). [Svelte tutorial uses a function named `forward`](https://svelte.dev/tutorial/event-forwarding), maybe that's a better name for that function?

app/modules/users/helpers.js
maxlath: renamed and refactored in 3b671b342

maxlath:
> when i load page `/users/network`, then i click on my username (aka got to my inventory `users/adamsberg/inventory`) my inventory is blank with no error message in console

Is that still the case will the latest version?

app/modules/users/components/users_home_layout.svelte
maxlath: renamed in 156136a

jum-s:
> > when i load page `/users/network`, then i click on my username (aka got to my inventory `users/adamsberg/inventory`) my inventory is blank with no error message in console
> 
> @jum-s Is that still the case will the latest version?

Yes.

Also, as you can see in the gif above, on the network page, a "There is nothing here" is displayed, while its still loading items. Would make more sense to have a spinner instead. If you don't fix it now (although it would be preferable), can we have an issue on the matter?

maxlath:
@jum-s that would be quite a regression if some inventories can not be displayed, so we have to solve it before merging. Anything in the server or client logs? Could you try to investigate where that fails?

jum-s:
Non-blocking/open issue was refering to the too early "there is nothing here" instead of spinner thinggy.

There is nothing in the console. Can you reproduce the bug? It only bugs when loading the page from `/users/network` (press F5) then go to your inventory. I can take a look a it tomo, despite not being in context.

maxlath:
@jum-s nice, I was able to reproduce the issue \o/

jum-s:
Looks like `users_home_layout` dont have `user` defined, thus not falling into any of the if categories line 66

maxlath:
> Looks like `users_home_layout` dont have `user` defined, thus not falling into any of the if categories line 66

yes, fixed in 098b3056b

jum-s:
for the spinner, it happens when loading (F5) the page `users/network`. Before items are loadedshould show a spinner instead of now "There is nothing here". After looking a bit at it, it look like `PaginatedItems` (triggered by NetworkUsersNav`) dont have any promise `waiting` to resolve on.

maxlath:
> for the spinner, it happens when loading (F5) the page `users/network`. Before items are loaded, it should show a spinner, instead of now a "There is nothing here". After looking a bit at it, it looks like `PaginatedItems` (triggered by `NetworkUsersNav`) dont have any promise `waiting` to resolve on.

should have been fixed by 1c78b7305

PR #399, opened 2023-03-29
Task svelte rewrite

jum-s:
This is a simple rewrite in svelte. Simple as in it doesnt touch/refactor `getNextTask` (only debackbonify it in 639336580bf263ea8f1cf2e8e13ca817be2f6c95).
Got for inspiration from DeduplicateWorks UI (without the selectable part).
The main chunk is in 2093b7886d9787e4e49ad798c7dc374e20886fcc which does some polymorphism (display work tasks and human tasks) ended up having a some `{#if entityType ...}` but nothing unbearable for the moment (but we can clearly feel it may become messy if some other task type are introduced later).
The only innovation is to sort first and highlight works that have a matched label between `from` & `to` entities and the source occurrence (cf. efbb7d6eae6f9d7ad2dc9e6b94bd347a4ae26cc0).


maxlath:
The rebase on master with autosquash has conflicts

PR #402, opened 2023-04-18
Create and edit instance of

jum-s:
This is a newer implementation post editor svelte.

not much innovation compared to #218, except for not displaying any input for P31s (aka `notSearchableProps`) to not give false impression that the value is editable/searchable.


AutomCoding:
This should probably be added to the [user rights table](https://wiki.inventaire.io/wiki/Roles_and_access_levels#Who_can_do_what_about_entities) on the wiki when merged.

PR #410, opened 2023-05-18
add missing images: update validation on an images object

jum-s:
Necessary with server PR [#685](https://github.com/inventaire/inventaire/pull/685), as former `isNonEmptyPlainObject(entityImages)` would otherwise always be true


app/modules/entities/lib/types/work_alt.js
```diff
@@ -31,7 +31,8 @@ async function addMissingImages (entities) {
   const imagesByUri = await getEntitiesImages(uris)
   entities.forEach(entity => {
     const entityImages = imagesByUri[entity.uri]
-    if (isNonEmptyPlainObject(entityImages)) setEntityImages(entity, entityImages)
+    const imagesUrls = extractImagesUrls(entityImages)
+    if (isNonEmptyArray(imagesUrls)) setEntityImages(entity, entityImages)
```
maxlath, line 35: what about just dropping the guard?
```suggestion
    setEntityImages(entity, entityImages)
```
The difference is that if there are no image, `entity.images` will equal `[]` rather than `undefined`

PR #412, opened 2023-06-07
add admin-only "latest users" layout

### LINKED ISSUE #238: create an admin-only layout to see the latest users created
To be done after https://github.com/inventaire/inventaire/issues/455

That layout could live on the route `/admin/users`. There is no `/admin/*` for now, so that would require creating a new module (see the [settings module](https://github.com/inventaire/inventaire-client/blob/d0c420a/app/modules/settings/settings.js) for an example of simple module)

maxlath:
goes with https://github.com/inventaire/inventaire/pull/688

Fixes #238

### DIFF
--- a/app/modules/users/components/latest_users.svelte
+++ b/app/modules/users/components/latest_users.svelte
@@ -25,7 +25,7 @@
   fetchMore()
 </script>
 
-<h2>{i18n('Latest users')}</h2>
+<h2>{i18n('Latest created users')}</h2>
 
 <Flash state={flash} />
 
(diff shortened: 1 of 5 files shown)


jum-s (approved):

app/modules/users/components/latest_users.svelte
```diff
@@ -0,0 +1,114 @@
...
+      users = users.concat(newUsers.map(serializeUser))
+    } catch (err) {
+      flash = err
+    }
+  }
+
+  fetchMore()
+</script>
+
+<h2>{i18n('Latest users')}</h2>
```
jum-s, line 28:
```suggestion
<h2>{i18n('Latest created users')}</h2>
```
[non-blocking]

Issue #202, opened 2020-04-12
Possibility to edit illustrators


thewholelifetolearn:
As of inventaire/inventaire#252, illustrators are linked to works.
At the moment, it is possible to see them when a work is displayed, but they can't be modified. As a side effect, many illustrators are designed as authors.

So it would be great to have the possibility to edit those information as well.

maxlath:
yes, I was thinking about that too: we could add illustrators (P110), but also colorists (P6338). Do you think it's ok to keep scenarist as authors (P50)?

My only concerns there are:
* the multiplication of patches: one patch to remove the P50 value, one to re-add that value as P110. It would be great if we could make both operations in a single patch
* we probably need a new gadget for that, to move a claim from one property to another easily, and, indeed, in a single patch

thewholelifetolearn:
It depends about Inventaire's philosophy to be (or not) a mirror of Wikidata database. If it is not the case, we can make the "shortcut" that a scenarist is an author too.

In the case of illustrators (and maybe colorists too), I have realized that this information is at the right place in Wikidata for the series. So it may be possible for works that are attached to a series to check if an author is listed as an illustrator is the series or not. If it is the case, then move the author as an illustrator.

In a long term, it would be good to have a functionality to move a property's value to another.

aazae:
Hi,

Linked to this issue, the cleanup tool for advanced users is mileading : 
![image](https://user-images.githubusercontent.com/88883564/143769385-cf5a631a-0643-4ddf-823f-ed26f64f3215.png)

If a series of work have the proper authors /illustrators / colorists property in wikidata, it is well displayed in the series view. However, in the cleanup tool, every illustrators and colorists are seen as "authors", and there is a prompt to add the missing authors in the individual works of the series.
The problem is that it add them as authors wrongly when doing so.

It should either :
- Not show them as authors in this view
- Add them with their correct properties in the individual works

(thx for this project, I love it :) )

PR #434, opened 2023-09-04
[WIP] add alternative author roles properties

maxlath:
addressing #202

### DIFF
--- a/app/modules/entities/components/editor/entity_edit.svelte
+++ b/app/modules/entities/components/editor/entity_edit.svelte
@@ -43,11 +43,11 @@
   </div>
 
   {#if !hasMonolingualTitle}
-    <LabelsEditor {entity} bind:favoriteLabel />
+    <LabelsEditor bind:entity bind:favoriteLabel />
   {/if}
 
   {#each Object.entries(typePropertiesPerCategory) as [ category, categoryProperties ]}
-    <PropertyCategory {entity} {category} {categoryProperties} />
+    <PropertyCategory bind:entity {category} {categoryProperties} />
   {/each}
 
   <div class="next">
--- a/app/modules/entities/components/editor/entity_edit.svelte
+++ b/app/modules/entities/components/editor/entity_edit.svelte
@@ -2,29 +2,32 @@
   import app from '#app/app'
   import { i18n, I18n } from '#user/lib/i18n'
   import LabelsEditor from './labels_editor.svelte'
-  import { propertiesPerTypeAndCategory } from '#entities/lib/editor/properties_per_type'
+  import { getTypePropertiesPerCategory } from '#entities/lib/editor/properties_per_type'
   import getBestLangValue from '#entities/lib/get_best_lang_value'
   import { loadInternalLink, icon } from '#lib/utils'
   import EntityEditMenu from './entity_edit_menu.svelte'
   import PropertyCategory from '#entities/components/editor/property_category.svelte'
   import { typesPossessiveForms } from '#entities/lib/types/entities_types'
+  import { onChange } from '#lib/svelte/svelte'
 
   export let entity
 
   const { uri, type, labels } = entity
-  const typePropertiesPerCategory = propertiesPerTypeAndCategory[type]
-  const hasMonolingualTitle = typePropertiesPerCategory.general['wdt:P1476'] != null
   const goToEntityPageLabel = `Go to the ${typesPossessiveForms[type]} page`
 
-  let favoriteLabel
+  let typePropertiesPerCategory, hasMonolingualTitle, favoriteLabel
 
-  $: {
+  function onEntityChange () {
+    typePropertiesPerCategory = getTypePropertiesPerCategory(entity)
+    hasMonolingualTitle = typePropertiesPerCategory.general['wdt:P1476'] != null
     if (hasMonolingualTitle) {
       favoriteLabel = entity.claims['wdt:P1476']?.[0]
     } else {
       favoriteLabel = getBestLangValue(app.user.lang, null, labels).value
     }
   }
+
+  $: onChange(entity, onEntityChange)
 </script>
 
 <div class="entity-edit">
@@ -42,13 +45,15 @@
     <EntityEditMenu {entity} />
   </div>
 
-  {#if !hasMonolingualTitle}
+  {#if hasMonolingualTitle === false}
     <LabelsEditor bind:entity bind:favoriteLabel />
   {/if}
 
-  {#each Object.entries(typePropertiesPerCategory) as [ category, categoryProperties ]}
-    <PropertyCategory bind:entity {category} {categoryProperties} />
-  {/each}
+  {#if typePropertiesPerCategory}
+    {#each Object.entries(typePropertiesPerCategory) as [ category, categoryProperties ]}
+      <PropertyCategory bind:entity {category} {categoryProperties} />
+    {/each}
+  {/if}
 
   <div class="next">
     <a
--- a/app/modules/entities/components/editor/select_author_role.svelte
+++ b/app/modules/entities/components/editor/select_author_role.svelte
@@ -7,9 +7,6 @@
 
   export let entity, property, value
 
-  // TODO: set rules to determine if a role change is possible, based on:
-  // - value is already in target role property claims
-
   let rolesProperties
   function setRolesProperties () {
     rolesProperties = getWorkPreferredAuthorRolesProperties(entity)
--- a/app/modules/entities/components/editor/select_author_role.svelte
+++ b/app/modules/entities/components/editor/select_author_role.svelte
@@ -1,4 +1,5 @@
 <script>
+  import { getWorkPreferredAuthorRolesProperties } from '#entities/lib/editor/properties_per_subtype'
   import preq from '#lib/preq'
   import { onChange } from '#lib/svelte/svelte'
   import { I18n, i18n } from '#user/lib/i18n'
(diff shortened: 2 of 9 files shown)


nclm:
Hello, just chipping in to make sure that the terminology is the right one :)

A comic book doesn’t usually have an “illustrator”, it has an “artist” (“dessinateur”⋅ice in French) and a “writer” (“scénariste” in French). Both of them are commonly called the “authors” of the book.

It would be incorrect to use `P110` for them. Checking Wikidata for various comic books, it is indeed `P50` that is used for both artists and writers. There are some comic books related properties such as letterer (`P9191`), inker (`P10836`), penciller (`P10837`), colorist (`P6338`), but I believe we are meant to use `P50` for artists and writers.

“Illustrators” are people making illustrations alongside a text, such as in some novels for instance, and other kind of works where the illustrations are not part of the main text. So the books that have “illustrators” are rarely or never comic books.

Moreover, sometimes there are illustrators that are just for an edition of a work: often, a classic text might be published many times with a different illustrator each time (and some editions without illustrations). Some other times, there might be a main illustrator that worked alongside the author and their illustrations are in all or most editions.

maxlath:
`P110` use to be the recommended property for "dessinateur⋅ice" (as shown by "dessinateur" being one of `P110` French aliases), possibly with a qualifier, but there is now indeed [`P10837`](https://www.wikidata.org/wiki/Wikidata:Property_proposal/penciller) that was created for that exact purpose, so that seems to be the way to go :)

nclm:
Comics (often US) only mention a “penciller” if there is also a separate “inker”. I don’t think we should use “penciller” for most comics, which have the same artist doing both.

It’s also not completely correct to use “illustrateur⋅ice/dessinateur⋅ice” for the artist if there is no special “scénariste” role for the writer. Both are “authors”, so it isn’t valid to mark only the writer as so. Note that there is also an growing movement to consider the colorist as one of the authors.

Just want to make sure we use the right semantic properties :) I do think this PR is a good idea, great to have the choice of the author role! I have been missing “illustrator” for several books, but especially for specific editions of novels.

jum-s:
When one create a new work, one choose Type: Comics, then P50 author slot disappear. 
But when one create a new work, one fill the author P50 slot, one choose Type: Comics `P31:Q1004`, then P50 author slot is still on.

It is understandable but slightly confusing to remove the author P50 slot. Maybe i dont have precise information about whose role the author exactly have, but i would like to have the possibility to still fill the author P50 slot (?)

jum-s:
When one create a new work, one fill the author P50 slot, one choose Type: Comics P31:Q1004, one changes from author to screenwriter with the little menu below the statement value
Then an error is thrown:
```
[Cannot have duplicate keys in a keyed each: Keys at index 1 and 5 with value 'wdt:P58' are duplicates
validate_each_keys@webpack-internal:///./node_modules/svelte/src/runtime/internal/each.js:147:10
update@webpack-internal:///./app/modules/entities/components/editor/entity_create.svelte:502:72
update@webpack-internal:///./app/modules/entities/components/editor/entity_create.svelte:839:42
update@webpack-internal:///./app/modules/entities/components/editor/entity_create.svelte:432:15
update@webpack-internal:///./app/modules/entities/components/editor/entity_create.svelte:253:16
update@webpack-internal:///./app/modules/entities/components/editor/entity_create.svelte:944:16
update@webpack-internal:///./node_modules/svelte/src/runtime/internal/scheduler.js:133:30
flush@webpack-internal:///./node_modules/svelte/src/runtime/internal/scheduler.js:93:11
]
```

jum-s:
Small feature suggestion: when an entity is "wrongly" associated with a work type (ie. a screenwriter for a written work), today if one moves it to author, then the screenwriter disappears. Which is fine if the user knows they are cleaning up the entity. But if they're not, it would be nice to suggest (ie. through a flash/infobox) that there is something smelly on this entity (à la Wikidata, with the tiny flag of the values constraints). It would be a nice little feature useful to this PR, that could then be generalized slowly on other cases.

[non-blocking, but to move in an issue]

app/modules/entities/components/editor/lib/suggestions/property_values_shortlist.js
```diff
@@ -1,18 +1,13 @@
-import preq from '#lib/preq'
+import { API } from '#app/api/api'
 import { pluralize } from '#entities/lib/types/entities_types'
+import assert_ from '#lib/assert_types'
+import preq from '#lib/preq'
+
+export const allowedValuesPerTypePerProperty = await preq.get(API.data.propertyValues).then(({ values }) => values)
```
jum-s, line 6: Wow, that's quite something to read. I guess that happens often to old fools who see the world (and webpack) moving around them.

jum-s:
When i edit an existing "written work" with an author and an screenwriter, aka with claims as such:
```
  "claims": {
    "wdt:P31": [
      "wd:Q47461344"
    ],
    "wdt:P58": [
      "inv:0380fed910ff9a4bb8cc319031cd264a"
    ],
    "wdt:P50": [
      "inv:0835b4d395020e8753e296937cb0b4bc"
    ]
  },
``` 
Then i change P31 to Comics (`wd:Q1004`), then i change the author role to screenwriter. 
An error says 
```
PossiblyUnhandledRejection: entity.claims[currentRoleProperty] is undefined

onRolePropertyChange@webpack-internal:///./app/modules/entities/components/editor/select_author_role.svelte:293:60
onChange@webpack-internal:///./app/lib/svelte/svelte.js:37:3
instance/$$self.$$.update@webpack-internal:///./app/modules/entities/components/editor/select_author_role.svelte:440:65
```

app/modules/entities/components/editor/lib/suggestions/property_values_shortlist.js
maxlath: Right?!? ^^ It has more to do with the [ES modules having been standardized as async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#top_level_await), than webpack which just complies to that standard

maxlath:
(Answering in review threads would be more convenient)

> `Cannot have duplicate keys in a keyed each`

should have been fixed by 9fb244a44

> It is understandable but slightly confusing to remove the author P50 slot

The idea is to encourage to directly enter the specific role. I can think of 2 alternatives, not sure which one is best:
- keeping the P50 slot but disabled, with a message encouraging to use more specific roles
- replacing the generic property editors by an even more ad-hoc "authors" editor, with a list of all contributors and the possibility to add their roles in place, that could look something like this : 

B

maxlath:
@nclm after 2db5845fe we now would have the following properties:

```js
const mostlyTextWorksRoles = [
  'wdt:P50', // author
  'wdt:P110', // illustrator
]
const drawedWorksRoles = [
  'wdt:P58', // scenarist
  'wdt:P6338', // colorist
  'wdt:P9191', // letterer
  'wdt:P10836', // inker
  'wdt:P10837', // penciller
]
```
Any change suggestion?

app/modules/entities/components/editor/select_author_role.svelte
```diff
@@ -0,0 +1,78 @@
...
+{#if rolesProperties && isNonEmptyClaimValue(value)}
+  <label>
+    {i18n('Change author role')}
+    <select bind:value={currentRoleProperty}>
+      {#each rolesProperties as roleProperty}
+        {@const disabled = roleProperty !== property && entity.claims[roleProperty]?.includes(value)}
+        <option
+          value={roleProperty}
+          {disabled}
+          title={disabled ? i18n('This author already has that role') : ''}
```
jum-s, line 61: `author` string can be ambiguous here, since the `role` can have `author` value. Then, when you hover on disabled `author` role, it feels redundant, one may say "This author already has that author role"). We could find a way to be more abstract: 

```suggestion
          title={disabled ? i18n('This person already has that role') : ''}
```
but are organizations persons?

Even more abstract:
```suggestion
          title={disabled ? i18n('They already has that role') : ''}
```

jum-s:
After selecting a new P31 value, it looks like image update is blinking:

Peek 2023-11-06 10-14

Looks like its a general problem (update new author is doing the same), therefore:

[non-blocking]

but, posting this comment in case you are in the editor context, and have an easy fix at hand, and feel like fixing...

jum-s:
> Answering in review threads would be more convenient

Yeah, this happens when i dont have located a specific line to address the comment. Duly noted though.

app/modules/entities/components/editor/select_author_role.svelte
maxlath: Depending on the choice made between the [atlernatives above](https://github.com/inventaire/inventaire-client/pull/434#issuecomment-1791275470), we might have the possibility to keep "author" without it adding confusion

app/modules/entities/components/editor/select_author_role.svelte
```diff
@@ -0,0 +1,78 @@
...
+    if (rolesProperties.length === 1 && rolesProperties[0] === property) {
+      rolesProperties = null
+    }
+  }
+
+  $: onChange(entity, setRolesProperties)
+
+  let currentRoleProperty = property
+
+  async function onRolePropertyChange () {
```
jum-s, line 26: not sure `onRolePropertyChange` is the culprit, but when editing [this entity](https://inventaire.io/entity/inv:035a93cc360f4e285e955bc12c282f10/edit), when i select first author "change author role" field to illustrator, then P50 author value (Anne Barber) successfully appear in illustrator prop field, but keeps appearing in author P50 field (and error is thrown if i want to select author P5 again `ERROR: claim property value not found`)

app/modules/entities/components/editor/entity_edit.svelte
```diff
@@ -42,13 +45,15 @@
...
-    <LabelsEditor {entity} bind:favoriteLabel />
+  {#if hasMonolingualTitle === false}
+    <LabelsEditor bind:entity bind:favoriteLabel />
   {/if}
 
-  {#each Object.entries(typePropertiesPerCategory) as [ category, categoryProperties ]}
-    <PropertyCategory {entity} {category} {categoryProperties} />
-  {/each}
+  {#if typePropertiesPerCategory}
+    {#each Object.entries(typePropertiesPerCategory) as [ category, categoryProperties ]}
```
jum-s, line 53: When editing [this entity](https://inventaire.io/entity/inv:00f118336b02219e1bddc8fa930e64be/edit), when i select P31 Comics, then change first author to inker, author is duplicated in Original Language field

app/modules/entities/components/editor/select_author_role.svelte
maxlath: Change done in 03ebb42a8. I think that we will have cases where author is still used as the generic term, while still covering more specific roles, for instance in the case of the Author layout, and I think it's ok

app/modules/entities/components/editor/entity_edit.svelte
maxlath: Should have been fixed by 228941f0e

app/modules/entities/components/editor/select_author_role.svelte
maxlath: Should have been fixed by 228941f0e

Issue #54, opened 2015-12-31
allow to pick position from geocoding instead of map marker


maxlath:
Options:
- OpenStreetMap geocoding API Nominatim http://nominatim.openstreetmap.org/search?q=schoneweide%20Berlin&format=json
- Google geocoding API https://maps.googleapis.com/maps/api/geocode/json?address=Schoneweide%20Berlin
- OpenCage Geocoder https://geocoder.opencagedata.com/ (requires an API key)
- [MapZen](https://mapzen.com/) (not recommended by @nerdyprojects )

maxlath:
this could also be useful to answer the problem stated in this feedback:
> is it possible to just type in a or town in a field when adding at town? I don't think everyone wants to allow the geolocation feature on their device or computer

maxlath:
* https://community.algolia.com/places/

maxlath:
* [GeoNames](http://consbio.github.io/Leaflet.Geonames/) (see also [Leaflet.Geonames](https://github.com/consbio/Leaflet.Geonames))

maxlath:
Done in https://github.com/inventaire/inventaire-client/pull/441

PR #441, opened 2023-09-29
allow to search a location by name

maxlath:
Addressing #54 with @antalAkos's help. This has been so long in the waiting, so excited to finally get there! :joy: :tada: 
It should help #295 and more generally having an easy to navigate public map.

TODO post-deploy:
- [ ] remove users and groups positions strictly equal to the [placeholder coordinates](https://github.com/inventaire/inventaire-client/blob/c9cb895bd4c625ea9f23a9b64cbddfa3c7a4383d/app/modules/map/lib/navigator_position.js#L33-L34)

### DIFF
--- a/app/modules/map/components/location_search_input.svelte
+++ b/app/modules/map/components/location_search_input.svelte
@@ -1,4 +1,5 @@
 <script>
+  import Spinner from '#components/spinner.svelte'
   import Flash from '#lib/components/flash.svelte'
   import { fitResultBbox, searchLocationByText } from '#map/lib/nominatim'
   import { i18n } from '#user/lib/i18n'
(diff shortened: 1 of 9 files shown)


app/modules/map/components/location_search_input.svelte
```diff
@@ -0,0 +1,117 @@
...
+  input{
+    margin: 0;
+  }
+  label{
+    font-size: 1rem;
+    margin-block-end: 0.2rem;
+  }
+  ul{
+    background-color: $off-white;
+    max-height: 20em;
```
jum-s, line 87:
```suggestion
    max-height: 20em;
    /* Small screens */
    @media screen and (max-width: $small-screen){
      max-height: 16em;
    }
```
to not overflow leaflet's attribution stripe

maxlath: addressed in edd45f0a8

maxlath:
[placeholder coordinates](https://github.com/inventaire/inventaire-client/blob/c9cb895bd4c625ea9f23a9b64cbddfa3c7a4383d/app/modules/map/lib/navigator_position.js#L33-L34) where removed will the following [data transform](https://github.com/inventaire/inventaire/blob/main/docs/couchdb_data_transformations.md) function, applied to both users and groups:

```js
// drop_placeholder_position.mjs
export default function (doc) {
  const { position } = doc
  // Take into account the different number of decimal used since this placeholder was introduced
  if (position[0].toString().startsWith('46.2324') && position[1].toString().startsWith('6.045')) {
    delete doc.position
  }
  return doc
}
```

PR #443, opened 2023-10-06
[WIP] Ordering listing elements

jum-s:
See server branch [#677](https://github.com/inventaire/inventaire/pull/677)


jum-s:
To old, will reboot later with different naming and mechanism.

PR #449, opened 2023-10-12
fix #448: Display book count in user profile

### LINKED ISSUE #448: Display book count in user profile
I only see the number of shelves at the top neart my profile picture. In the code, it seems there is some provision to display the number of books.

lasconic:


lasconic:
Anything I need to do for this PR to be accepted ?

jum-s:
After some digging, it appears that this book count display on user profile was existing but [it has not been re-implemented by the Svelte rewrite](https://github.com/inventaire/inventaire-client/blob/master/app/modules/users/components/user_profile.svelte#L21), only the [template](https://github.com/inventaire/inventaire-client/blob/master/app/modules/users/components/user_profile.svelte#L53) has been rewritten. I think it would be preferable to find out why `inventoryLength` is not populated anymore.
You are welcome to rebase your attempts on this pull request or close this and open a new one. Thanks !

lasconic:
> I think it would be preferable to find out why inventoryLength is not populated anymore.

I believe this PR does just that. It populates inventoryLength in main_user.js

jum-s:
Yes, it does populate the main user. I was thinking of having a broader solution to display inventoryLength on all other user profiles (as issue title suggest). But it does solve what got raised in the matrix channel. 
Having a `inventoryLength` on every user profile can come later, at Svelte rewrite . 
Thanks!

lasconic:
For all users, it's more involved indeed since I'm not sure which 'length' we want to display. Private/public books? Books the main user could see ?

PR #456, opened 2023-11-11
Dont display authors strings if an author entity exists

jum-s:
Do not display author strings (P2093), when a work already has some authors (P50), to keep some information on poor entities (ie. articles). This will remove author strings as they are lower quality than P50 graph data values. 

This is a low hanging fruit, as author strings duplicates could be cleaned up on Wikidata, but, as @maxlath explored, it would requires quite some work as references would need some care (ie. [Q5460282](https://www.wikidata.org/wiki/Q5460282)).


maxlath:
I think author strings create confusion as that means there are now two types of authors, with different behaviors (different destinations when you click on them, not appearing in the inventory browser, not appear in the editor where you expect to see authors, etc). I would be in favor of fully ignoring them for everything but articles (and even there it stinks imho). There might be a possibility to use those author strings as suggestions in the editor, but it seems also to be potentially complex/confusing.

jum-s:
> and even there it stinks imho

All right, it got completely removed even for articles in [93cf3d4](https://github.com/inventaire/inventaire-client/pull/456/commits/93cf3d4a4294304902fe9aceabbd6168ae1ac824)

PR #463, opened 2024-01-08
Entity editor: add links to properties documentation

maxlath:
It could be useful to contributors who want to learn more about properties format, constraints, data modeling considerations,  etc

### DIFF
--- a/app/modules/entities/components/editor/property_claims_editor.svelte
+++ b/app/modules/entities/components/editor/property_claims_editor.svelte
@@ -124,7 +124,7 @@
     font-weight: normal;
     @include display-flex(row, center, space-between);
     &:first-child{
-      margin-block-start: 0.2em;
+      margin-block-start: 0;
     }
     &:not(:first-child){
       margin-block-start: 1em;
--- a/app/modules/entities/components/editor/property_claims_editor.svelte
+++ b/app/modules/entities/components/editor/property_claims_editor.svelte
@@ -9,7 +9,7 @@
   import { getPropertyClaimsCount, isEmptyClaimValue, isNonEmptyClaimValue } from '#entities/components/editor/lib/editors_helpers'
   import Flash from '#lib/components/flash.svelte'
   import { isWikidataPropertyUri } from '#lib/boolean_tests'
-  import { getWikidataPropertyUrl } from '#lib/wikimedia/wikidata'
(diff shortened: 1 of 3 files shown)


app/modules/entities/components/editor/property_claims_editor.svelte
```diff
@@ -66,7 +68,21 @@
     {#if isRequiredAndMissing}
       <span class="required-notice">{I18n('required')}</span>
     {/if}
-    <h3 class="editor-section-header">{I18n(customLabel || property)}</h3>
+    <div class="property-info">
+      <h3 class="editor-section-header">{I18n(customLabel || property)}</h3>
+      {#if isWikidataPropertyUri(property)}
+        <a
+          class="uri"
+          href={getWikidataPropertyDocumentationUrl(property)}
```
jum-s, line 76: Would it make sense to restrict this display to dataadmin ? to not have the additional ~15-20% vertical space on empty values, for an information that might clutter the UI for users comfortable on the inv editor (and dont care about WD).
Alternatives: 
- make the property title as a link itself (like it happens already to entity label on editor page for example).
- add the link inline, after the property title

Im a bit concerned about liberally adding information to the UI, as we should be careful of keeping it easy to read (this is also a note for future self).

jum-s: actually tweaking some CSS to make it more compact seems enough with 7b0df8292

maxlath: I wonder if we could also just show that info to users who have associated a wikidata account to there user account

jum-s: neat, this is a usecase that could fit as a new role `wikidatauser` which would be updated after oauth autorisation/revocation

maxlath: I don't think we need a new `role` for that, that's just inferred from the state of `user.oauth.wikidata`. That could also be a user display setting that is either automatically turned on when connecting a wikidata account, or offered to be turned on at that moment(?)

jum-s: A "Display setting" sounds good, as some users might want access to Wikidata documentation without OAuth-ing they're user account.

PR #464, opened 2024-01-12
entities layouts:  ease external ids links customization

maxlath:
As we now have plenty of links from external ids to display, having good interfaces to customize what should be displayed seems important. In particular, a user that doesn't explore the settings might not realize that they can display all those links.

### DIFF
--- a/app/modules/entities/components/layouts/category_external_ids.svelte
+++ b/app/modules/entities/components/layouts/category_external_ids.svelte
@@ -4,6 +4,7 @@
   import { user } from '#user/user_store'
   import { I18n, i18n } from '#user/lib/i18n'
   import { icon } from '#lib/utils'
+  import { getTextDirection } from '#lib/active_languages'
 
   export let category, categoryAvailableExternalIds
 
@@ -17,7 +18,7 @@
   $: displayedCategoryExternalIds = showAllAvailableExternalIds ? categoryAvailableExternalIds : categoryPreferredAvailableExternalIds
 </script>
 
-<p class="category">
+<p class="category" dir={getTextDirection($user?.language)}>
   <span class="category-label">{I18n(categoryLabels[category])}:</span>
   {#each displayedCategoryExternalIds as { property, name, value }, i}
     <EntityClaimLink {property} {name} {value} />{#if i !== displayedCategoryExternalIds.length - 1},&nbsp;{/if}
@@ -46,6 +47,11 @@
     :global(.fa){
       font-size: 0.8rem;
       width: 0.7rem;
+    }
+  }
+  [dir="rtl"]{
+    button :global(.fa){
+      transform: rotate(180deg);
     }
   }
 </style>
--- a/app/modules/entities/components/layouts/category_external_ids.svelte
+++ b/app/modules/entities/components/layouts/category_external_ids.svelte
@@ -12,6 +12,7 @@
 
   let showAllAvailableExternalIds = false
   let showCategorySettings = false
+  const linksId = `${category}Links`
 
   let categoryPreferredAvailableExternalIds, displayedCategoryExternalIds
   $: {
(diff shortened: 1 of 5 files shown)


app/modules/settings/components/displayed_links.svelte
```diff
@@ -34,45 +42,86 @@
...
-      <input type="checkbox" bind:group={selectedWebsites} value={websiteName} />
-      {websiteName}
+{#if category == null || category === 'bibliographicDatabases'}
+  <section>
+    <label class="category">
+      <input
+        type="checkbox"
+        checked={selectedBibliographicDatabasesCount === bibliographicDatabasesNames.length}
+        indeterminate={selectedBibliographicDatabasesCount !== 0 && selectedBibliographicDatabasesCount !== bibliographicDatabasesNames.length}
+        on:click={() => toggleAllSection({ names: bibliographicDatabasesNames, selectedCount: selectedBibliographicDatabasesCount })}
```
jum-s, line 67: I recall a conversation where you where against this checkbox-that-modify-everything behavior. The reason behind was that it is not very accessible, as what is going to happen after clicking on the checkbox is left unknown to the user. Usecase: one crafts their ids customization, then discovers the general checkbox, and on their first click on it, they deletes their own customisation for good.

`Select All` button pattern seems to be the accepted solution since inventory browser implementation. And is now in other parts of the app (ie. items map, entity homonym selections). 


So, i think it would make sense, on a less central display than inventory browser (where space is less compact/important), to reuse this pattern, wouldnt it?

app/modules/entities/components/layouts/category_external_ids.svelte
```diff
@@ -0,0 +1,91 @@
...
+            {i18n('display_x_links', { smart_count: categoryAvailableExternalIds.length })}
+          {:else}
+            {i18n('display_x_more', { smart_count: categoryAvailableExternalIds.length - categoryPreferredAvailableExternalIds.length })}
+          {/if}
+        {/if}
+      </button>
+    {/if}
+    <button title={i18n('Customize which links should be displayed')} on:click={() => showCategorySettings = true}>
+      {@html icon('cog')}
+    </button>
```
jum-s, line 47: What about displaying the cog icon only after clicking on "display more", to keep default display as simple as possible ?

maxlath: Done in 9f2258ccf. I was afraid it might create a bit of a hide and sick, with cases where it's impossible to access the cog button, but it those make the default display heavy with the cog.

app/modules/settings/components/displayed_links.svelte
jum-s: > it respects a known convention

I couln't find this convention anywhere in the codebase though. On the contrary the "select all" pattern appears in more than the inventory browser (ie. map filters)

I would still be more comfortable with an already in use pattern (and more accessible (cf. my last comment)) to not have different display with same behavior.

If you want to continue with this unnamed checkbox, can the `input` element be extracted into its own component (to make reusable for map filters) ?

maxlath: > I couln't find this convention anywhere in the codebase though

I meant a known convention in general, but that has indeed not been used so far in the codebase

maxlath: This [implementation with the button and a dynamic label](https://www.sncf-voyageurs.com/fr/voyagez-avec-nous/horaires-et-itineraires/itineraires/) seems to work nicely

jum-s: I dont know how to continue this, i think its a mistake to present two different ways to display the same behavior. And sure, its minimalistic, but also more difficult to understand whats going on. This app is already complicated enough to make the user guess what will happen if checking the box.

You seem definitive, so i will swallow it, can we at least extract it in its own component to reuse it elsewhere ?

maxlath: I tried to see how to extract the input, but that was not conclusive: that looks like a case of painful-to-DRY-so-lets-copy.

maxlath:
Merging before this further rots (the rebase was painful), but let's see what the user feedback is.

PR #474, opened 2024-02-09
Article table view

jum-s:
More compact view for articles sections (+ 2 details related to article layout)

Inspired by scholia's display:

Tested entities: 
  - author layout: `entity/wd:Q63799298`, `/entity/wd:Q37638169`
  - article layout: `/entity/wd:Q104739517`

Adding columns for P123 and/or P921 (aka publisher, main subject) would require some additional section typing to fetch appropriate related entities. This may be done later, in anther PR.

### DIFF
--- a/app/modules/entities/components/layouts/article.svelte
+++ b/app/modules/entities/components/layouts/article.svelte
@@ -1,11 +1,11 @@
 <script>
+  import EntityTitle from './entity_title.svelte'
   import { i18n } from '#user/lib/i18n'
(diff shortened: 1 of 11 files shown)


maxlath (approved):
Yep, that's better! If we went to add more data in the future, we might need to toggle to column mode on small screens. I stole that idea somewhere for Time Capsule, and I think it works quite well:
* Large screen
a
* Small screen
c

But that can wait I think

PR #482, opened 2024-02-22
entity_layout_actions: add 'refresh external databases data' button on edition layouts

maxlath:
to trigger the server behavior introduced by https://github.com/inventaire/inventaire/pull/734

### DIFF
--- a/app/modules/entities/components/layouts/entity_layout_actions.svelte
+++ b/app/modules/entities/components/layouts/entity_layout_actions.svelte
@@ -7,13 +7,12 @@
   import Spinner from '#general/components/spinner.svelte'
   import app from '#app/app'
   import { tick } from 'svelte'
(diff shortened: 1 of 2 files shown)


app/modules/entities/components/layouts/entity_layout_actions.svelte
```diff
@@ -72,6 +73,21 @@
     />
   </li>
 {:else}
+  {#if isIsbnEntityUri(uri)}
```
jum-s, line 76: It's an old issue (https://github.com/inventaire/inventaire/issues/192) but since removing isbn's uris is still a possibility, what about directly checking if isbn exists ? (Since server's resolver is using this prop value, while uri is an indirection)
```suggestion
  {#if claims['wdt:P212']}
```

maxlath: done in c88a487

PR #485, opened 2024-03-22
entity_merge: add "view result" link below the flash message

### LINKED ISSUE #478: Link or redirect to entity after a merge
After merging an item into another, Inventaire shows this screen:

image

This is nice, but how do I get to my merged entity to continue editing? I have to press the back button at least 3 or 4 times, and that somehow does not always work.

So we could imagine having a link, or even redirecting, to the merged entity after the merge has been performed.

maxlath:
I just merged #481, but maybe I closed it to fast: I'm not fully certain it's enough to close #478, as clicking the autocomplete field with the preserved target entity would not get you to that page, but instead display the text input to change the entity value. So this PR proposes to display a more explicit link.

### DIFF
--- a/app/modules/entities/components/entity_merge.svelte
+++ b/app/modules/entities/components/entity_merge.svelte
@@ -8,10 +8,11 @@
   import { isWikidataItemUri } from '#lib/boolean_tests'
   import { slide } from 'svelte/transition'
   import Spinner from '#components/spinner.svelte'
+  import Link from '#lib/components/link.svelte'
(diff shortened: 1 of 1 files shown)


jum-s (approved):
I think it's still possible to access `to` entity from autocomplete field (on both large and mobile screens), but a link makes it more straight forward indeed.
Todo/Food for thought: add a `<slot/>` in `flash.svelte` to display this kind of link content.

PR #492, opened 2024-05-01
editor: add label removal possibilities

jum-s:
While this feature is implemented by doing some duck typing a bit (aka depending on whether `value` is defined as `updateOrRemoveLabel`, it will update or remove label accordingly), but the cases are a bit complex if you want to add the delete button (it would simplify a lot not to have it, but it does follow claim editor mechanism).

### DIFF
--- a/app/api/entities.ts
+++ b/app/api/entities.ts
@@ -75,6 +75,7 @@ export default {
 
   labels: {
     update: action('update-label'),
+    remove: action('remove-label'),
   },
 
   merge: action('merge'),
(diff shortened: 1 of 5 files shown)


maxlath (commented):
I implemented the undo button in [eccf388](https://github.com/inventaire/inventaire-client/pull/492/commits/eccf388a58eca4cab6ec544183f99d8a27a07743), can you give it a try?

jum-s:
As undo feature adds some complexity, I added e4a2bfa2c to fix some difficult states one may end up in ie. add a new label, then delete another one, then select an other language should return a wrong labels list, such as:

PR #491, opened 2024-04-22
Ordering listing elements

jum-s:
Synced with server branch: [677](https://github.com/inventaire/inventaire/pull/677)

This branch allows reorder any list elements. All the lexicographical ordering logic is handled server side.
This naive client requests the new ordinal position it would like to see the element in the current list. Therefore, requests parameters should look like: 

```
{
  id: 'a8a18f31fc5148dd46ac80e2b70c7a91',
  ordinal: 2,
  reqUserId: 'a8a18f31fc5148dd46ac80e2b70ad099'
}
```

PR #501, opened 2024-06-26
Listing element comment

jum-s:
This improves listing display of each element.
Comment can now be written to an element.

Should be rebased and merged after #491 is merged on main.

Goes with server PR https://github.com/inventaire/inventaire/pull/751

### Detailed specifications 
from past meeting with max

- have a confirmation modal when unchecking checkbox of button "add to..."

- open URL `lists/aaa/element/:id` should redirect to modal `lists/:id/element/:id` 

- closing modal should go back on list page

- edit element: button cote a cote (a la item ou shelf editor)

- summary should appear below comment in modal

- add work image in modal

- have serif on titles


jum-s:
Note that having two distinct modals (a "view modal" to display information and a "pencil modal" to edit an element) feels a bit clogged, aka it could make sense to not have the "pencil modal" and instead to edit an element's comment directly in the view modal. 
**But**, pencil modal will be improved soon (moving an element to another list (which is a milestone)), and opens for improvments on the long run too (ie. moving an element to a defined position in the list, enabling element visibility)

maxlath:
UI feedback: I click on the pencil to add a comment, then save and close the modal. The comment doesn't appear anywhere. I then click on the element's body, which displays the modal, nothing either. (=> element state propagation is missing)

After reloading the page, the comment does appear. I click on the element's body again, which displays the info modal with my comment: there is now no direct way to edit that comment; I have to know that I need to close the modal to open the edit modal. => I still have strong feelings against having two different modals. I think there should be only one modal, just like for items: the element owner would just be given additional edit options.

app/modules/listings/components/listing_element.svelte
```diff
@@ -23,31 +33,90 @@
...
+      app.navigate(getListingPathname(listingId), options)
+    }
+  }
+
+  $: comment = element.comment
 </script>
-<div class="listing-element-section">
+{#if isShowMode}
+  <Modal on:closeModal={toggleShowMode}
+  >
```
maxlath, line 52: The modal is unnecessarily narrow on large screens, especially noticeable when there is a lot of info
```suggestion
  <Modal on:closeModal={toggleShowMode} size="large">
```

app/modules/listings/components/listing_element_actions.svelte
```diff
@@ -45,16 +52,28 @@
         elements = oldElements
       })
   }
+
+  function toggleEditMode () {
+    isEditMode = !isEditMode
+  }
 </script>
+{#if isEditMode}
```
maxlath, line 60: [non-blocking]
```suggestion
{#if showEditModal}
```

maxlath (changes requested):
UI feedback: I click on the pencil to add a comment, then save and close the modal. The comment doesn't appear anywhere. I then click on the element's body, which displays the modal, nothing either. (=> element state propagation is missing)

After reloading the page, the comment does appear. I click on the element's body again, which displays the info modal with my comment: there is now no direct way to edit that comment; I have to know that I need to close the modal to open the edit modal. => I still have strong feelings against having two different modals. I think there should be only one modal, just like for items: the element owner would just be given additional edit options.

app/modules/listings/components/listing_element.svelte
jum-s: done in d3d025a7c

jum-s: actually, i think it needs a more refined size customization
Small screens looks too narrow:

app/modules/listings/components/listing_element_actions.svelte
jum-s: done in https://github.com/inventaire/inventaire-client/commit/d3d025a7cce87e7747df00df53771d57ebf2367b

app/modules/listings/components/listing_element.svelte
maxlath: I wish we had a linter rule for this, but using `:global` at the component root should be forbidden, as it's then unscopped and can thus leak to affect other parts of the app. My alternative proposal: revert fd9de6e and pick ac42be86c from the branch [`listing-element-comment-max`](https://github.com/inventaire/inventaire-client/tree/listing-element-comment-max)

app/modules/listings/components/listing_element.svelte
```diff
@@ -23,70 +29,130 @@
   } else if (image?.url) {
     imageUrl = image.url
   }
+
+  function toggleShowMode () {
+    showElementModal = !showElementModal
+    autocompleteFlash = null
+    if (showElementModal) {
+      const options = { metadata: getElementMetadata(listing, element), preventScrollTop: true }
```
maxlath, line 37: This line crashes when after editing the comment I close the modal and try to re-open it:

```
entity is undefined
getElementTitle@webpack-internal:///./app/modules/listings/lib/listings.ts:212:13
async*getElementMetadata@webpack-internal:///./app/modules/listings/lib/listings.ts:184:18
toggleShowMode@webpack-internal:///./app/modules/listings/components/listing_element.svelte:688:90
```

app/modules/listings/components/listing_element_comment.svelte
```diff
@@ -0,0 +1,121 @@
...
+        class="tiny-button validate success"
+        title={I18n('validate')}
+        on:click={validate}
+      >
+        {@html icon('check')}
+        {I18n('validate')}
+        {#await validating}
+          <Spinner />
+        {/await}
+      </button>
```
maxlath, line 72: [non-blocking]
```suggestion
      <button
        class="tiny-button validate success"
        title={I18n('validate')}
        disabled={validating != null}
        on:click={validate}
      >
        {#if validating}
          <Spinner />
        {:else}
          {@html icon('check')}
        {/if}
        {I18n('validate')}
      </button>
```

Two things there:
- Show the spinner where the icon is to prevent an unesthetical width jump
- Disable the button while validating to prevent a double request. This requires being able to use `validating` as a boolean, which can be done either by doing `validating = null` once the promise resolved, or by turning `validating` into a boolean. I tend to prefer this boolean approach these days, which is a reversal on an earlier opinion of mine, sorry for that :sweat_smile:

app/modules/listings/components/listing_element_comment.svelte
```diff
@@ -0,0 +1,121 @@
...
+
+  function toggleEditMode () {
+    isEditMode = !isEditMode
+    flash = null
+  }
+
+  $: comment = element.comment
+  async function _updateElement () {
+    flash = null
+    try {
```
maxlath, line 36: [non-blocking]
That would be the perfect place to insert the spam filter introduced by 22a77a8c9
```suggestion
    try {
      await checkSpamContent(newComment)
```

app/modules/listings/components/listing_element.svelte
jum-s: nicely spotted, this was a mistake. I pushed a fixup commit 99a01ff03 reusing the rules i defined earlier since i would like to keep the `40em` since its the size of the lisitng elements in the background. And also prefering widely existing rules (@media queries) than your proposal.

app/modules/listings/components/listing_element_comment.svelte
jum-s: > I tend to prefer this boolean approach these days

Any reason behind this reversal? I still think the boolean is more "manipulable"

addressed in 80b3f45ff

app/modules/listings/components/listing_element_comment.svelte
jum-s: Yes, lets add it once its rebased or after merge to not rewrite all commits ids in a rebase

app/modules/listings/components/listing_element.svelte
jum-s: fixed in ce6f68597

app/modules/listings/components/listing_element.svelte
maxlath: I'm not really convinced, but let's give it a shot like this and see if it stays itchy

app/modules/listings/components/listing_element_comment.svelte
maxlath: The main point for preferring booleans IMO is the ability to pass it to the `disabled` attribute: getting a boolean from a promise is surprisingly not straight-forward (I wish `Promise.isResolved` was a thing, but it isn't)

app/modules/listings/components/listing_element_show.svelte
```diff
@@ -0,0 +1,125 @@
...
+
+  <Summary {entity} />
+
+  {#if element.comment || isCreatorMainUser}
+    <ListingElementComment
+      {isCreatorMainUser}
+      bind:element
+      bind:flash
+    />
+  {/if}
```
maxlath, line 57: [non-blocking, could be adjusted later]
I think it would make sense to display the comment just below the title, as, when it exists, it's the most important part of that page. I think that's already true when opening the element modal from the list context, but will be even more when opening it from the entity context: if you want to read what different users have to say about the entity, you are not primarily interested in reading the entity details over and over

app/modules/listings/components/listing_element_show.svelte
```diff
@@ -0,0 +1,125 @@
...
+        size={256}
+      />
+    {/if}
+    <div class="entity-infobox">
+      <AuthorsInfo {claims} />
+      <Infobox
+        {claims}
+        entityType={type}
+        shortlistOnly={true}
+      />
```
maxlath, line 44: Authors info are displayed twice on works/series. Example here with a list element based on wd:Q8337

app/modules/listings/lib/listings.ts
```diff
@@ -87,8 +102,51 @@ export async function getListingMetadata (listing) {
...
+  const { name, creator } = listing
+  const { username } = await app.request('get:user:data', creator)
+  return `${entity.title} - ${name} - ${i18n('list_created_by', { username })}`
+}
+
+export async function askUserConfirmationAndRemove (removeElementPromise, deletingData) {
+  if (deletingData) {
+    app.execute('ask:confirmation', {
+      // i18n suggestion:  "delete_element_confirmation": "Are you sure you want to delete this element. Choosing yes will delete the element and the following data:  \"**%{deletingData}**\"?",
+      confirmationText: i18n('delete_element_confirmation', { deletingData }),
```
maxlath, line 130: Underscore keys aren't needed anymore to do string interpolation
```suggestion
      confirmationText: i18n('Are you sure you want to **delete this element**? That will also delete your comment: %{comment}…', { comment: comment.slice(0, 20) }),
```

maxlath (changes requested):

app/modules/listings/components/listing_element_comment.svelte
jum-s: since it got rebased, done in adad16576

app/modules/listings/components/listing_element_comment.svelte
```diff
@@ -0,0 +1,128 @@
...
+</div>
+<style lang="scss">
+  @import "#general/scss/utils";
+  .edit-button{
+    @include display-flex(row, flex-start, space-between);
+    width: 100%;
+<<<<<<< Updated upstream
+=======
+    @include shy(0.5);
+>>>>>>> Stashed changes
```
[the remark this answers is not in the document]
jum-s: done in 0a2c1a6a2

app/modules/listings/components/listing_element_show.svelte
jum-s: Attempt done in b21c68acf

app/modules/listings/components/listing_element_comment.svelte
```diff
@@ -0,0 +1,128 @@
...
+      <button
+        class="tiny-button validate success"
+        title={I18n('validate')}
+        disabled={validating}
+        on:click={validate}
+      >
+        {#if validating}
+          <Spinner />
+        {:else}
+          {@html icon('check')}
```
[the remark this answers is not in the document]
jum-s: done in b21c68acf

app/modules/listings/components/listing_element_show.svelte
jum-s: done in f942b90f9

PR #502, opened 2024-06-28
Listing entities type

jum-s:
Creates list with a specific entity type only (works, publishers).

If no type is passed when creating the list, `work` is assigned by default.
Series entities maybe added to a work listing.
Adding an element by having a scoped search according to listing entities' type.

Solves issue https://github.com/inventaire/inventaire/issues/716
Server PR [#752](https://github.com/inventaire/inventaire/pull/752)
I18n PR [12](https://github.com/inventaire/inventaire-i18n/pull/12)


app/modules/listings/components/listing_editor.svelte
```diff
@@ -99,10 +107,25 @@
...
-  bind:visibility
-  showTip={true}
-/>
+<div class="visibility-selector">
+  <VisibilitySelector
+    bind:visibility
+    showTip={true}
+  />
+</div>
+<label>
```
maxlath, line 116: I think the listing type selector should be placed before the name input field, as that's really determining the nature of the list and should thus be considered first by users creating a list

app/modules/listings/lib/search_types.ts
```diff
@@ -0,0 +1,27 @@
+import { typesBySection } from '#search/lib/search_sections'
```
maxlath, line 1: This file seems to be an obsolete duplicate of `app/modules/listings/lib/entities_typing.ts`

maxlath (changes requested):

maxlath (approved):

PR #503, opened 2024-07-01
Task reorganise duplicates

jum-s:
Goes with server side https://github.com/inventaire/inventaire/pull/735

running server's tasks test should create some tasks fixtures for client testing
Tasks URLs to tests:
- `/tasks/humans` 
- `/tasks/works`


maxlath:
Rebased and dropped the commits belonging to https://github.com/inventaire/inventaire-client/pull/504

app/modules/entities/components/layouts/authors_info_role.svelte
```diff
@@ -18,7 +18,7 @@
           { @const hasManyClaimValues = claims[prop].length > 15}
           {#each claims[prop] as claimValue}
             <AuthorDisplay
-              entity={authorsByUris[claimValue]}
+              {authorsByUris}
```
maxlath, line 21: I think that's a mistake: the component doesn't actually need the other entities data. I suspect you did that to work around an issue with components failing to update on changes? Doesn't using keyed each, something like ` {#each claims[prop] as claimValue (claimValue)}` fix the issue?

jum-s: yes, done in 4bed90370

maxlath: Then can this line be reverted to `entity={authorsByUris[claimValue]}`?

jum-s: i dont understand why it hasnt been pushed actually, heres another try in 23cb87de5

PR #506, opened 2024-08-09
Scanner rewrite in Svelte

maxlath:
Rewriting `/add/scan` at feature parity. Taking the opportunity to replace [`quagga`](https://github.com/serratus/quaggaJS) with [`quagga2`](https://github.com/ericblade/quagga2), which is maintained and comes with types.


app/modules/inventory/lib/scanner/quagga.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const verticalMargin = '30%'
+const horizontalMargin = '15%'
+
+// See doc: https://github.com/ericblade/quagga2#quaggainitconfig-callback
+function getConfig (containerElement: HTMLElement) {
+  return {
+    inputStream: {
+      name: 'Live',
+      type: 'LiveStream' as InputStreamType,
```
jum-s, line 71: `LiveStream` is the [default value](https://github.com/ericblade/quagga2?tab=readme-ov-file#inputstream) can then be removed  
```suggestion
```

app/modules/inventory/lib/scanner/quagga.ts
```diff
@@ -0,0 +1,104 @@
...
+      },
+      target: containerElement,
+      constraints: getConstraints(),
+    },
+    // From the docs https://github.com/ericblade/quagga2?tab=readme-ov-file#locate:
+    // "Why would someone turn this feature off? Localizing a barcode is a computationally expensive operation and might not work properly on some devices.
+    // Another reason would be the lack of auto-focus producing blurry images which makes the localization feature very unstable.
+    // However, even if none of the above apply, there is one more case where it might be useful to disable locate: If the orientation, and/or the approximate position of the barcode is known,
+    // or if you want to guide the user through a rectangular outline. This can increase performance and robustness at the same time."
+    locate: false,
```
jum-s, line 86: Some config improvement: default frequency is 10 image analysed per second, which would save some CPU is lowered to 5 (even 2 looks good to me)
```suggestion
    locate: false,
    frequency: 5
```

jum-s, line 86: Some config improvement: from the doc: `In cases where the barcode is further away from the camera lens (lack of auto-focus, or small barcodes) then it's advised to set the size to small`
Book barcodes a generally quite small, ad some cameras cannot focus enough if pu to close (to fill the green box), i think we should give it a shot. When tested locally, it did improve how many barcodes were recognized 
```suggestion
    locate: false,
    locator: {
      patchSize: "small"
    }
```

app/modules/inventory/lib/scanner/quagga.ts
maxlath: done in f9a8214e3

app/modules/inventory/lib/scanner/quagga.ts
maxlath: This makes close scans fail, so that would require decreasing the green box to let users have a clue on the smaller expected size, or a way to detect if the camera can autofocus. I would prefer to leave that kind of adjustments for later.

PR #513, opened 2024-10-08
Display coauthors in entityListRow

jum-s:
Which is a feature parity, since before Svelte transition. It was not implemented at the time of the entities' layout rewriting because pooling another level of entities fetching felt like too much work at the time.
Now that things have calm down, this introduces a `coauthors` line in a work row (that is on an author layout for example).

To do in another PR: extend the behaviour to author's series


jum-s:
Closing this for now, as it demands too much typing and too much refactoring of `getWorksAuthorsUris` and we dont have much time now

Issue #432, opened 2023-08-16
Allow to move a listing element to another listing


maxlath:
My current use-case: I would like to move an element from "reading" to "read in 2023"

PR #523, opened 2024-11-20
Move element to another list

jum-s:
Provides a dropdown on element modal in order to move an element to another list, which will also move any data attached to it.

Server PR https://github.com/inventaire/inventaire/pull/795
Solves issue https://github.com/inventaire/inventaire-client/issues/432

To be merged after #502

### DIFF
--- a/app/modules/listings/components/listing_element_show.svelte
+++ b/app/modules/listings/components/listing_element_show.svelte
@@ -15,7 +15,7 @@
   import Infobox from '#entities/components/layouts/infobox.svelte'
   import Summary from '#entities/components/layouts/summary.svelte'
   import { authorsProps } from '#entities/components/lib/claims_helpers'
-  import { getEntitiesTypes } from '#listings/lib/entities_typing'
+  import { entitiesTypesByListingType } from '#listings/lib/entities_typing'
   import { updateElement } from '#listings/lib/listings'
   import { userListings } from '#listings/lib/stores/user_listings'
   import ListingElementComment from '#modules/listings/components/listing_element_comment.svelte'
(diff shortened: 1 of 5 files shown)


app/modules/listings/components/listing_element_show.svelte
```diff
@@ -68,8 +98,43 @@
...
+                {#each recipientListings as recipientListing}
+                  <button
+                    role="menuitem"
+                    on:click={() => updateElementListing(recipientListing._id)}
+                    tabindex="-1"
+                  >
+                    {recipientListing.name}
+                  </button>
+                {/each}
+              </div>
```
maxlath, line 131: [non-blocking]
This kind of dropdown menu should ideally be keyboard accessible
```suggestion
              <ul role="menu" use:autofocus>
                {#each recipientListings as recipientListing}
                  <li>
                    <button
                      role="menuitem"
                      on:click={() => updateElementListing(recipientListing._id)}
                    >
                      {recipientListing.name}
                    </button>
                  </li>
                {/each}
              </ul>
```
With
```js
  import { autofocus } from '#app/lib/components/actions/autofocus'
```

app/modules/listings/components/listing_element_show.svelte
```diff
@@ -105,24 +170,39 @@
...
+    &:hover, &:focus{
       background-color: $danger-color;
-      color: white;
     }
   }
+  [role="menu"] button{
+    @include shy-border;
+    @include tiny-button-padding;
+    @include bg-hover(white, 5%);
+    padding: 1em;
```
maxlath, line 201: [non-blocking]
That produces big buttons; what about giving them the same aspect as when displayed within the dot dot dot menu (minus the checkbox)?

maxlath (approved):

app/modules/listings/components/listing_element_show.svelte
jum-s: done in 75ae9a946 significantly smaller than last version, but i think the dotdotdot menu items are too small 75ae9a946

app/modules/listings/components/listing_element_show.svelte
jum-s: Definitely, well spotted: done in 530463774

PR #526, opened 2024-12-09
Add the possibility to pool activities together

jum-s:
Server branch : [#802](https://github.com/inventaire/inventaire/pull/801)

### DIFF
--- a/app/modules/settings/components/account.svelte
+++ b/app/modules/settings/components/account.svelte
@@ -13,8 +13,9 @@
   import { user } from '#user/user_store'
   import EmailValidation from './email_validation.svelte'
 
-  let flashLang, flashEmail, flashFediversable
+  let flashLang, flashEmail, flashDiscoverability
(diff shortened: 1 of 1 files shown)


app/modules/settings/components/account.svelte
```diff
@@ -151,7 +166,17 @@
         on:click={toggleFediversable} />
       {i18n('Fediverse integration')}
     </label>
-    <Flash bind:state={flashFediversable} />
+    <!-- I18n copy/pasted from fediversable_description last paragraph: To prevent sending too many messages, when several books are added at about the same time, only one summary message will be sent. -->
+    <p class="note">{@html I18n('pooling_activities_description')}</p>
```
maxlath, line 170: We don't need underscored i18n keys anymore
```suggestion
    <p class="note">{i18n('To prevent sending too many messages, when several books are added at about the same time, only one summary message will be sent.')}</p>
```

PR #538, opened 2025-03-30
Finish transition to Svelte

maxlath:
* Rewrites the last views in Svelte: `notifications_layout` (+ children), `confirmation_modal`, and finally (:tada:) `app_layout` (+ children)
* Replaces uses of handlebars `escapeExpression` by [`escape-html`](https://www.npmjs.com/package/escape-html), to be able to remove `handlebars` as a dependency
* Purges obsolete code :fire: 

This PR touches many files, and as we don't have automated tests for the client, manually testing all around the app is welcome

Remaining traces of Backbone/Marionette will be removed in #539

PR #539, opened 2025-04-02
Remove last bits of Backbone, Marionette, and jQuery

maxlath:
Internalizes minimal implementations of a router, and event bus, mimicking respectively `Backbone.Router` and `backbone.radio`.

`Backbone.history` is replaced by the native [History API](https://developer.mozilla.org/en-US/docs/Web/API/History_API)

~~Depends on #538~~


app/init_app_layout.ts
```diff
@@ -0,0 +1,36 @@
+import AppLayout from '#components/app_layout.svelte'
+import { commands } from './radio'
+import type { SvelteComponent, ComponentProps, ComponentType } from 'svelte'
+
+export let appLayout
```
jum-s, line 5: Loosening dependencies, love it!

jum-s:
when I click on "Add edition without ISBN", an error is shown: 

```
Error: TypeError: expected string, got undefined (undefined)
    newError webpack-internal:///./app/lib/error.ts:20
    assertType webpack-internal:///./app/lib/assert_types.ts:26
    assertString webpack-internal:///./app/lib/assert_types.ts:70
    showEntityCreateFromRoute webpack-internal:///./app/modules/entities/entities.ts:135
    loadUrl webpack-internal:///./app/lib/router.ts:51
    navigate webpack-internal:///./app/app.ts:72
    navigateAndLoad webpack-internal:///./app/app.ts:85
    loadInternalLink webpack-internal:///./app/lib/utils.ts:129
```

maxlath:
> when I click on "Add edition without ISBN", an error is shown

Fixed by f2a5664

maxlath:
> When I am on a list with element, I click on an element, then it glimpses and reload the whole list's page to open element modal. This behaviour is not happening on main.

Fixed on main by 3af75aeb5 (has it was caused by de69eaa3, which is already on main)

>  if I hit previous to go back to an element URL (lists/:id/element/:elId), nothing happen

Fixed by 5eb8db279

maxlath:
Merging as all reported problems have been fixed and this PR fixes known bugs currently in prod. But of course, if more issues are identified, they can be reported and addressed here.

